PR #3, opened 2016-11-22
Added models

kamijean:


lib/blacklight.rb
```diff
@@ -1,9 +1,81 @@
...
+require_relative "./blacklight/exceptions"
 
 module Blacklight
+  def self.parse(args)
+    source_directory = validates_source_directory(args[0])
+    output_directory = args[1]
+    opens_dir(source_directory, output_directory)
+  end
+
+  def self.validates_source_directory(directory)
```
bfcoder, line 19: I still think this could be named better. :]
```ruby
def self.verifed_source_directory
  ...
end
```
or
```ruby
def self.validated_source_directory
  ...
end
```

bfcoder (commented):
In the middle of reviewing this PR and it was merged. Someone needs to pass off the code in a PR before it is merged.

PR #5, opened 2016-12-02
Quizlet

kamijean:
Adding importation of quizzes


codingfoo (commented):
Remove ExportFile_Bb

lib/blacklight/models/answer.rb
```diff
@@ -0,0 +1,30 @@
...
+    attr_reader :id
+    attr_accessor :fraction, :resp_ident, :feedback
+
+    def initialize(text, id = nil)
+      @answer_text = text
+      @resp_ident = ""
+      @fraction = ""
+      @feedback = ""
+      @points = 0
+      @id = id ? id : Blacklight.create_random_hex
```
bfcoder: would
```ruby
@id = id || Blacklight.create_random_hex
```
work just as well?

lib/blacklight/models/assessment.rb
```diff
@@ -0,0 +1,95 @@
...
+      assessment.points_possible = @points_possible
+      assessment = create_items(assessment)
+      assessment.assignment = assignment
+      assessment
+    end
+
+    def create_items(assessment)
+      items = []
+      @items = @items - ["", nil]
+      @items.each do |item|
```
bfcoder: consider
```ruby
items = @items.map do |item|
  Question.from(item).
    iterate_xml(item)
end
```

Also, would it make sense to store the `item` in the Question class so you don't have to pass it to `#iterate_xml`?

lib/blacklight/models/assessment.rb
```diff
@@ -0,0 +1,95 @@
...
+      assessment.items = []
+      items.each do |item|
+        assessment = item.canvas_conversion(assessment)
+      end
+      assessment
+    end
+
+    def create_assignment_group(course)
+      group = course.assignment_groups.detect { |a| a.title == @group_name }
+      if group == nil
```
bfcoder: consider swapping these?
```ruby
if group
  @group_id = group.identifier
else
  ...
end
```

lib/blacklight/models/question.rb
```diff
@@ -0,0 +1,199 @@
...
+        @incorrect_answers["fraction"].to_f
+      end
+    end
+
+    def set_answers(resprocessing)
+      set_correct_answers(resprocessing)
+      set_incorrect_answers(resprocessing)
+    end
+
+    def set_correct_answers(resprocessing)
```
bfcoder, line 118: This and `#set_incorrect_answers` are almost identical. Let's DRY this up and use one method?

lib/blacklight/models/question.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    def iterate_item(data)
+      @general_correct_feedback = set_correct_feedback(data)
+      @general_incorrect_feedback = set_incorrect_feedback(data)
+      @material = set_material(data)
+      resprocessing = data.children.at("resprocessing")
+      @max_score = set_max_score(resprocessing)
+    end
+
+    def set_correct_feedback(data)
```
bfcoder: Same here for feedback. Let's DRY it up?

lib/blacklight/models/question.rb
```diff
@@ -0,0 +1,199 @@
...
+        correct_feedback.search("mat_formattedtext").text
+      else
+        ""
+      end
+    end
+
+    def set_incorrect_feedback(data)
+      incorrect_feedback = data.children.
+        css("itemfeedback[ident=incorrect]").first
+      incorrect_children = incorrect_feedback.children.at("mat_formattedtext")
```
bfcoder: Looks like this is done a bit different than above. Which is correct?

lib/blacklight/models/question.rb
```diff
@@ -0,0 +1,199 @@
...
+      else
+        ""
+      end
+    end
+
+    def set_max_score(resprocessing)
+      no_score = "0.0"
+      outcomes = resprocessing.search("outcomes")
+      if outcomes && !outcomes.search("decvar").empty?
+        if outcomes.search("decvar")[0].attributes["maxvalue"]
```
bfcoder: Let's avoid magic numbers that don't describe anything. Either try and do this without it, or set it to a constant with a meaningful name.
```ruby
OUTCOME = 0
outcomes.search("decvar")[OUTCOME].attributes["maxvalue"]
```
or something like that.

lib/blacklight/models/questions/fill_in_blank.rb
```diff
@@ -0,0 +1,12 @@
+module Blacklight
+  class FillInBlank < Question
+    def iterate_xml(data)
+      super
+      conditionvar = data.search("resprocessing")[0].search("conditionvar")[0]
```
bfcoder: `data.at("resprocessing").at("conditionvar")`

lib/blacklight/models/questions/fill_in_blank_plus.rb
```diff
@@ -0,0 +1,17 @@
+module Blacklight
+  class FillInBlankPlus < Question
+    def iterate_xml(data)
+      super
+      conditionvar = data.search("resprocessing")[0].search("conditionvar")[0]
```
bfcoder: `data.at("resprocessing").at("conditionvar")`

lib/blacklight/models/questions/jumbled_sentence.rb
```diff
@@ -0,0 +1,38 @@
...
+  class JumbledSentence < Question
+    def initialize
+      @responses = []
+      super
+    end
+
+    def iterate_xml(data)
+      super
+      if response_block = data.children.search("flow[@class=RESPONSE_BLOCK]")
+        choices = []
```
bfcoder, line 11: consider?
```ruby
choices = response_block.
  children.
  at("flow_label").
  children.
  map do |response|
    text = response.children.at("mattext").text
    { id: response.attributes["ident"], text: text }
  end
```

lib/blacklight/models/questions/matching.rb
```diff
@@ -0,0 +1,57 @@
...
+    def initialize
+      super
+      @matches = []
+      @matching_answers = {}
+    end
+
+    def iterate_xml(data)
+      super
+      matches_array = []
+      if match_block = data.search("flow[@class=RIGHT_MATCH_BLOCK]")[0]
```
bfcoder: Magic number

bfcoder: Maybe `.first`?

lib/blacklight/models/questions/matching.rb
```diff
@@ -0,0 +1,57 @@
...
+        match_block.children.each do |match|
+          match_text = match.children.at("mat_formattedtext").text
+          matches_array.push(convert_html(match_text))
+        end
+      end
+      if response_block = data.search("flow[@class=RESPONSE_BLOCK]")[0]
```
bfcoder: `.first`?

lib/blacklight/models/questions/numeric.rb
```diff
@@ -0,0 +1,35 @@
...
+  class Numeric < Question
+    def initialize
+      super
+      @ranges = {}
+      @tolerances = {}
+    end
+
+    def iterate_xml(data)
+      super
+      conditionvar = data.search("resprocessing")[0].search("conditionvar")[0]
```
bfcoder: `data.at("resprocessing").at("conditionvar")`

lib/blacklight/models/questions/numeric.rb
```diff
@@ -0,0 +1,35 @@
...
+      if conditionvar
+        max_difference = data.search("decvar")[0].attributes["maxnumber"].to_i
```
bfcoder: `data.at("decvar")` should be the same as `data.search("decvar")[0]`

lib/blacklight/models/questions/ordering.rb
```diff
@@ -0,0 +1,38 @@
...
+    def canvas_conversion(assessment)
+      super
+      @question.matches = @matches
+      assessment
+    end
+
+    def process_response(resprocessing)
+      super
+      correct = resprocessing.css("respcondition[title=correct]")
+      correct.search("and")[0].children.each_with_index do |varequal, index|
```
bfcoder: `correct.at("and")` should be the same as `correct.search("and")[0]`

lib/blacklight/models/question.rb
bfcoder: or
`outcomes.at("decvar")` is better

lib/blacklight/xml_parser.rb
```diff
@@ -46,19 +39,19 @@ def self.iterate_xml(resources, zip_file)
...
-      data = Nokogiri::XML.parse(data_file)
-      xml_data = data.children.first
-      type = xml_data.name.downcase
-      if RESOURCE_TYPE[type.to_sym]
-        resource_type = "Blacklight::" + RESOURCE_TYPE[type.to_sym]
-        res_class = resource_type.split("::").
-          reduce(Object) { |o, c| o.const_get c }
-        resource = res_class.new
-        resources_array.push(resource.iterate_xml(xml_data))
+      unless zip_file.find_entry(file_name) == nil
```
bfcoder: I try to not use `unless meh == nil`
consider `if zip_file.find_entry(file_name)` instead

lib/blacklight/models/assessment.rb
jamis: I think it makes sense to keep `item` out of `Question`, since `item`shouldn't be part of the question's persistent state, but it might make sense to have `Question.from` both instantiate the question, _and_ invoke `#iterate_xml` (in other words, have `Question.from` return the fully initialized Question object).

lib/blacklight/models/answer.rb
jamis: Another option:

```ruby
def initialize(text, id=Blacklight.create_random_hex)
   # ....
end
```

This is only effective if you're treating `id` solely as an optional parameter. If you're ever passing an actual `nil` value as the `id` parameter, and you want `nil` converted to something else, then do it like @bfcoder suggests instead.

lib/blacklight/models/assessment.rb
```diff
@@ -0,0 +1,95 @@
...
+module Blacklight
+  class Assessment
+    def initialize
+      @title = ""
+      @description = ""
+      @quiz_type = "assignment"
+      @points_possible = 0
+      @items = []
+      @group_name
+      @group_id
```
jamis: Looks like you might have forgotten to assign anything these two variables.

lib/blacklight/models/questions/either_or.rb
```diff
@@ -0,0 +1,31 @@
...
+  class EitherOr < Question
+    EITHER_OR = {
+      "yes_no.true" => "Yes",
+      "yes_no.false" => "No",
+      "agree_disagree.true" => "Agree",
+      "agree_disagree.false" => "Disagree",
+      "right_wrong.true" => "Right",
+      "right_wrong.false" => "Wrong",
+      "true_false.true" => "Yes",
+      "true_false.false" => "Yes",
```
jamis: Is this second "Yes" supposed to be a "No"?

lib/blacklight/models/questions/matching.rb
bfcoder: `.at_css` actually

lib/blacklight/models/questions/matching.rb
bfcoder: `.at_css` here I think

lib/blacklight/models/assessment.rb
bfcoder: I like that better @jamis 👍

bfcoder: We should still use a map here though.

lib/blacklight/models/assessment.rb
```diff
@@ -0,0 +1,95 @@
...
+      assessment.points_possible = @points_possible
+      assessment = create_items(assessment)
+      assessment.assignment = assignment
+      assessment
+    end
+
+    def create_items(assessment)
+      items = []
+      @items = @items - ["", nil]
+      @items.map do |item|
```
bfcoder:
```ruby
items = @items.map do |item|
  Question.from(item)
end
```
and remove `items = []`

lib/blacklight/models/questions/matching.rb
```diff
@@ -0,0 +1,60 @@
...
+      @matching_answers = {}
+    end
+
+    def iterate_xml(data)
+      super
+      resprocessing = data.at("resprocessing")
+      @matching_answers = set_matching_answers(resprocessing)
+      matches_array = []
+      if match_block = data.at("flow[@class=RIGHT_MATCH_BLOCK]")
+        match_block.children.each do |match|
```
bfcoder: use `.map`
I like to avoid pushing into an array when I can use a `.map`

bfcoder:
Code looks good

PR #6, opened 2016-12-05
adds file blacklist

nick-benoit14:
* Adds file blacklist so only desired files are uploaded

### DIFF
--- a/lib/blacklight/xml_parser.rb
+++ b/lib/blacklight/xml_parser.rb
@@ -36,6 +36,14 @@ module Blacklight
     # safeassign: :iterate_safeassign,
   }.freeze
 
+  FILE_BLACK_LIST = [
+    ".dat",
+    "glossary",
+    "imsmanifest.xml",
+    ".bb-package-info",
+    ".bb-package-sig",
+  ].freeze
+
   def self.parse_manifest(zip_file, manifest)
     doc = Nokogiri::XML.parse(manifest)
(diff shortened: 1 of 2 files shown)


lib/blacklight/xml_parser.rb
```diff
@@ -61,12 +69,19 @@ def self.iterate_xml(resources, zip_file)
     resources_array - ["", nil]
   end
 
+  def self.black_listed?(name)
```
bfcoder, line 72: Consider?
```ruby
def self.black_listed?(name)
  !FILE_BLACK_LIST.include? name
end
```

lib/blacklight/xml_parser.rb
```diff
@@ -61,12 +69,19 @@ def self.iterate_xml(resources, zip_file)
...
+      return false if name.include? entry
+    end
+    true
+  end
+
   def self.iterate_files(zip_file)
     resources_array = []
-    zip_file.entries.each do |entry|
+    filtered_entries = zip_file.entries.select { |e| black_listed? e.name }
+    filtered_entries.each do |entry|
```
bfcoder: Let's refactor this too.

```ruby
  def self.iterate_files(zip_file)
    zip_file.
      entries.
      select { |entry| black_listed? entry.name }.
      map { |entry| BlacklightFile.new(entry) }
  end
```

bfcoder (approved):
Looks great!

PR #7, opened 2016-12-06
Move rake tasks into lib to be bundled with gem

bfcoder:
Adjust some code to compensate for new location of tasks


kamijean (approved):
Looks good and better then it was

PR #8, opened 2016-12-08
Add directories to the file blacklist

kamijean:


nick-benoit14:
Looks good to me.

lib/blacklight/xml_parser.rb
```diff
@@ -61,14 +61,15 @@ def self.iterate_xml(resources, zip_file)
...
+    FILE_BLACK_LIST.any? { |black_item| File.fnmatch?(black_item, name) } ||
+      type == :directory
   end
 
-  def self.iterate_files(zip_file)
-    zip_file.
-      entries.select { |e| !black_listed?(e.name) }.
-      map { |entry| BlacklightFile.new(entry) }
+  def self.iterate_files(zipfile)
+    zipfile.entries.select { |e| !black_list?(e.name, e.ftype) }.map do |entry|
```
bfcoder: why this refactor? it looked much cleaner multilined the way it was

bfcoder: I think you should revert this back to the way it was and add the `e.ftype` as needed

lib/blacklight/xml_parser.rb
```diff
@@ -61,13 +61,14 @@ def self.iterate_xml(resources, zip_file)
...
+  def self.black_list?(name, type)
+    FILE_BLACK_LIST.any? { |black_item| File.fnmatch?(black_item, name) } ||
+      type == :directory
   end
 
-  def self.iterate_files(zip_file)
-    zip_file.
-      entries.select { |e| !black_listed?(e.name) }.
+  def self.iterate_files(zipfile)
+    zipfile.
```
bfcoder: Don't hate me, but one more thing. :] Can you move entries up to this line? Keep the select and map on separate lines?
```ruby
zipfile.entries.
  select {}.
  map {}
```
😁

lib/blacklight/xml_parser.rb
```diff
@@ -61,13 +61,14 @@ def self.iterate_xml(resources, zip_file)
     resources_array.flatten - ["", nil]
   end
 
-  def self.black_listed?(name)
-    FILE_BLACK_LIST.any? { |b_list_item| File.fnmatch?(b_list_item, name) }
+  def self.black_list?(name, type)
+    FILE_BLACK_LIST.any? { |black_item| File.fnmatch?(black_item, name) } ||
+      type == :directory
```
bfcoder, line 66: In the interest of performance, should we do this check first?

bfcoder: nevermind, it's fine the way it is.

bfcoder:
Looks good.

PR #9, opened 2016-12-09
Add ability to upload imscc files

bfcoder:
Create a course
Upload content and process
Add specs


lib/blacklight/canvas_course.rb
```diff
@@ -0,0 +1,66 @@
+require "pandarus"
```
nick-benoit14: Should this course be moved to the 'models' folder?

bfcoder:
No, because it isn't one of the models for the canvas structure. CanvasCourse doesn't make sense to me to be in there for that reason. It's not really a model.

nick-benoit14:
Makes sense. You can go ahead an merge then. Thanks James!

bfcoder:
Will you pull the branch down and test it too?

nick-benoit14:
The specs are passing, I do get a 302, which makes rest-client throw an exception, when it is trying to upload the package. I think I may not have something set up quite right. Did you run into that at all?

This is where it is trying to make the post request to `https://au.openlmshost.com/files_api`

bfcoder:
I don't get that exception. Which was weird because canvas said it would happen but it would never happen to me. The docs say it should send back a 302 and I would need to post to the url it came back with in the header as the "location" for it to start processing. But for me it always just start processing... So I need to dig in a little more it looks like.

bfcoder:
Or are you talking about in the specs you get that?

nick-benoit14:
Nope, the specs are fine. It is the post request that is being made by `RestClient` in upload_content.  That is odd though, I was working with it for a while yesterday, and I have been consistently been redirected.

conradbeach:
Looks good to me.

PR #10, opened 2016-12-12
Refactor imscc convert task to operate on one file at a time

bfcoder:
Only call parse on individual files
Remove spammy output


kamijean (approved):
Looks good 👍

PR #13, opened 2016-12-14
Add scorm support

nick-benoit14:
* Adds ability to extract scorm packages from blackboard export 
* Adds ability to upload scorm packages to scorm manager 'adhesion' 
* Adds ability to create assignments from uploaded scorm packages
* Authenticates scorm manager requests


blacklight.yml.example
```diff
@@ -0,0 +1,19 @@
+# Generally looks like https://< mycanvas_instance >/api
+:canvas_url: <canvas instance api url>
+
+# Canvas tokens can be generated at <my_canvas_url>/profile/setting provided
+# that the user has the required priviledges
+:canvas_token: <canvas token>
+
+# Url of scorm manager. In the case of AU this will be the url of the adhesion
```
bfcoder: Don't mention AU

lib/blacklight.rb
```diff
@@ -53,9 +54,10 @@ def self.create_canvas_course(resources, zip_name)
     course
   end
 
-  def self.initialize_course(filename)
-    metadata = Blacklight::CanvasCourse.metadata_from_file(filename)
-    course = Blacklight::CanvasCourse.from_metadata(metadata)
-    course.upload_content(filename)
+  def self.initialize_course(canvas_import, blackboard_export)
+    metadata = Blacklight::CanvasCourse.metadata_from_file(canvas_import)
```
bfcoder: It's not a canvas_import, it is the filename to the import. So filename is important here. Actually, file_path is more appropriate, as it will be `canvas/Export.imscc`

Consider?
```ruby
def self.initialize_course(canvas_file_path, blackboard_file_path)
  metadata = Blacklight::CanvasCourse.metadata_from_file(canvas_file_path)
  bb_zip = if blackboard_file_path
             Zip::File.read(blackboard_file_path)
           end
  course = Blacklight::CanvasCourse.from_metadata(metadata, bb_zip)
  course.upload_content(filename)
end
```

lib/blacklight.rb
```diff
@@ -53,9 +54,10 @@ def self.create_canvas_course(resources, zip_name)
...
+    bb_zip = Zip::File.new(blackboard_export) unless blackboard_export.nil?
```
bfcoder: Use `Zip::File.read()` so the file is closed automatically.

lib/blacklight/canvas_course.rb
```diff
@@ -11,9 +12,19 @@ class CanvasCourse
     # A new canvas course accepts the metadata for a course
     # and the pandarus course resourse
     ##
-    def initialize(metadata, course_resource)
+    def initialize(metadata, course_resource, blackboard_export)
       @metadata = metadata
       @course_resource = course_resource
+      @scorm_packages = self.class.get_scorm_packages(blackboard_export)
```
bfcoder: use `CanvasCourse` instead of self.class

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,70 @@ def self.from_metadata(metadata)
...
+    ##
+    def create_scorm_assignment(scorm_package, course_id)
+      payload = {
+        assignment: {
+          name: scorm_package["title"],
+          submission_types: ["external_tool"],
+          integration_id: scorm_package["package_id"],
+          integration_data: { provider: "atomic-scorm" },
+          external_tool_tag_attributes: {
+            url: "#{Blacklight.scorm_launch_url}" \
```
bfcoder: This will have a newline in the string.
Consider?
```ruby
launch_url = Blacklight.scorm_launch_url
url = "#{launch_url}?course_id=#{scorm_package['package_id']}"
payload = {
  ...
  url: url
  ...
}
```

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,70 @@ def self.from_metadata(metadata)
...
+              "?course_id=#{scorm_package['package_id']}",
+          },
+        },
+      }
+
+      RestClient.post(
```
bfcoder: Can you not use pandarus here? https://canvas.instructure.com/doc/api/assignments.html#method.assignments_api.create

```ruby
client.create_assignment(
  course_id,
  scorm_package["title"],
  payload,
)
```

blacklight.yml.example
```diff
@@ -0,0 +1,19 @@
+# Generally looks like https://< mycanvas_instance >/api
```
bfcoder: This YML example is great for development scenarios. However, this is a gem, where you won't have this example yml available. A more appropriate place would be to document this in the readme. You could document in the readme and also say see this example yml. But all configurations really should be in the readme.

blacklight.yml.example
bfcoder: Consider `Url of scorm manager. This could be an [adhesion](github.com/atomicjolt/adhesion) app.`

lib/blacklight/models/scorm_package.rb
```diff
@@ -0,0 +1,89 @@
+module Blacklight
+  class ScormPackage
+    attr_accessor(:entries, :manifest)
+
+    ##
+    # Scorm packages should include this string in the <schema> tag. We
+    # downcase, and remove spaces to be more consistant
+    ##
+    SCORM_SCHEMA = "adlscorm".freeze
```
bfcoder, line 10: I don't know what this is exactly, but is it something that could potentially be configurable and should be in the yml?

lib/blacklight/models/scorm_package.rb
```diff
@@ -0,0 +1,89 @@
...
+    ##
+    def write_zip(export_name)
+      @@dir ||= Dir.mktmpdir
+      scorm_path = File.dirname @manifest.name
+      path = "#{@@dir}/#{export_name}"
+      Zip::File.open path, Zip::File::CREATE do |zip|
+        @entries.each do |entry|
+          if entry.file?
+            zip.get_output_stream(
+              self.class.correct_path(entry.name, scorm_path),
```
bfcoder: use `ScormPackage` instead of `self.class`

spec/fixtures/scorm_package/Module1/index.htm
```diff
@@ -0,0 +1,123 @@
...
+<script src="../_sys/systemCheck.js" type="text/javascript"></script>
+<script src="navigation.js" type="text/javascript"></script>
+<script>
+
+var mm_adl_API = null;
+var startPage = "";
+
+function mm_getAPI()
+{	
+	var myAPI = null;
```
bfcoder: Fix the indentation throughout this file to 2 spaces

Gemfile
```diff
@@ -4,3 +4,5 @@ source "https://rubygems.org"
 gemspec
 
 gem "canvas_cc", git: "https://github.com/atomicjolt/canvas_cc.git"
+
+gem "rubocop", require: false
```
bfcoder: I personally don't think this is beneficial to have in the gemfile

spec/fixtures/scorm_package/Module1/index.htm
```diff
@@ -0,0 +1,123 @@
+<!doctype html>
+<html>
+<link rel="stylesheet" type="text/css" href="../_sys/systemCheck.css" />
```
bfcoder: Do we really need these link and script tags in this fixture?

bfcoder: Do we even need this file at all? Or just the manifest?

lib/blacklight.rb
nick-benoit14: Do you mean Zip::File.open()? It doesn't seem that read will do what I need it to do.

lib/blacklight/models/scorm_package.rb
nick-benoit14: It is part of the scorm spec that this is included in the manifest. I thought it belonged as a class constant instead of just being a string inside of the method.

spec/fixtures/scorm_package/Module1/index.htm
nick-benoit14: We need the file to test whether or not additional files are being correctly added. I started with an actual entire scorm package, then it was slow so I removed most of it. I can remove the extraneous html though.

Gemfile
nick-benoit14: This is a merging mistake

lib/blacklight.rb
```diff
@@ -53,9 +54,12 @@ def self.create_canvas_course(resources, zip_name)
     course
   end
 
-  def self.initialize_course(filename)
-    metadata = Blacklight::CanvasCourse.metadata_from_file(filename)
-    course = Blacklight::CanvasCourse.from_metadata(metadata)
-    course.upload_content(filename)
+  def self.initialize_course(canvas_file_path, blackboard_file_path)
```
nick-benoit14: make optional parameter

lib/blacklight/models/file.rb
```diff
@@ -38,6 +38,7 @@ def canvas_conversion(course)
     # Remove temporary files
     ##
     def self.cleanup
+      @@dir ||= nil
```
BenHSmith13, line 41: Woah, this looks crazy. I don't know what it is, but I like it.

lib/blacklight/models/scorm_package.rb
```diff
@@ -0,0 +1,90 @@
...
+    rescue Nokogiri::XML::XPath::SyntaxError
+      false
+    end
+
+    ##
+    # Returns array of all scorm manifest files inside of blackboard export
+    ##
+    def self.find_scorm_manifests(zip_file)
+      return [] if zip_file.nil?
+      zip_file.
```
BenHSmith13: Why is this a newline

spec/fixtures/scorm_package/Module1/index.htm
```diff
@@ -0,0 +1,6 @@
+<!doctype html>
+<html>
+<body>
+	<h1>Howdy</h1>
```
bfcoder: use 2 spaces instead of a tab

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,71 @@ def self.from_metadata(metadata)
...
+      )
+    end
+
+    ##
+    # Uploads a scorm package to scorm manager specified in blacklight.yml
+    # config file
+    ##
+    def upload_scorm_package(scorm_package, course_id, tmp_name)
+      zip = scorm_package.write_zip tmp_name
+      resp = RestClient.post(
```
bfcoder: remove resp

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,71 @@ def self.from_metadata(metadata)
...
+        "#{Blacklight.scorm_url}/api/scorm_courses",
+        {
+          oauth_consumer_key: "scorm-player",
+          lms_course_id: course_id,
+          file: File.new(zip, "rb"),
+        },
+        SharedAuthorization: Blacklight.scorm_shared_auth,
+      )
```
bfcoder:
```ruby
) do |resp|
  JSON.parse(resp.body)["response"]
end
```

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,71 @@ def self.from_metadata(metadata)
...
+          submission_types: ["external_tool"],
+          integration_id: scorm_package["package_id"],
+          integration_data: { provider: "atomic-scorm" },
+          external_tool_tag_attributes: {
+            url: url,
+          },
+        },
+      }
+
+      CanvasCourse.client.create_assignment(
```
bfcoder, line 90: You have to pass (course_id, assignment_title, opts) to this method. As it is now, you are passing that hash as the assignment name.

bfcoder: image

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,71 @@ def self.from_metadata(metadata)
...
+      RestClient.post(
+        "#{Blacklight.scorm_url}/api/scorm_courses",
+        {
+          oauth_consumer_key: "scorm-player",
+          lms_course_id: course_id,
+          file: File.new(zip, "rb"),
+        },
+        SharedAuthorization: Blacklight.scorm_shared_auth,
+      ) do |resp|
+        return JSON.parse(resp.body)["response"]
```
bfcoder: Remove the return here, it is unnecessary in ruby.

nick-benoit14: Will it return the value from my outer function implicitly?

bfcoder (approved):
Code looks good. Pulled down and tested and seems to work good too.

lib/blacklight.rb
```diff
@@ -53,9 +54,12 @@ def self.create_canvas_course(resources, zip_name)
     course
   end
 
-  def self.initialize_course(filename)
-    metadata = Blacklight::CanvasCourse.metadata_from_file(filename)
-    course = Blacklight::CanvasCourse.from_metadata(metadata)
-    course.upload_content(filename)
+  def self.initialize_course(canvas_file_path, blackboard_file_path = nil)
+    metadata = Blacklight::CanvasCourse.metadata_from_file(canvas_file_path)
+    bb_zip = if blackboard_file_path
```
jamis, line 59: A more idiomatic way of saying this would be:

```ruby
bb_zip = Zip::File.open(blackboard_file_path) if blackboard_file_path
```

lib/blacklight/canvas_course.rb
```diff
@@ -11,9 +12,19 @@ class CanvasCourse
...
+    def initialize(metadata, course_resource, blackboard_export)
       @metadata = metadata
       @course_resource = course_resource
+      @scorm_packages = CanvasCourse.get_scorm_packages(blackboard_export)
+    end
+
+    ##
+    # Extracts scorm packages from a blackboard export zip file
+    ##
+    def self.get_scorm_packages(blackboard_export)
```
jamis, line 24: I think this method ought to belong on `ScormPackage` directly, rather than as a class method on `CanvasCourse`, because (1) it returns an array of `ScormPackage` instances, and (2), it doesn't actually refer to anything that is specific to `CanvasCourse`.

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,72 @@ def self.from_metadata(metadata)
...
-      CanvasCourse.new(metadata, canvas_course)
+      CanvasCourse.new(metadata, canvas_course, blackboard_export)
+    end
+
+    ##
+    # Creates a canvas assignment from a scorm package that has already been
+    # uploaded to a scorm manager
+    ##
+    def create_scorm_assignment(scorm_package, course_id)
+      url = "#{Blacklight.scorm_launch_url}?" +
```
jamis, line 76: Consider putting the question mark in the second string, and referencing `Blacklight.scorm_launch_url` directly, instead of via string interpolation, like this:

```ruby
url = Blacklight.scorm_launch_url +
        "?course_id=#{scorm_package['package_id']}"
```
Maybe that would even make the line short enough that it doesn't need to wrap.

lib/blacklight/canvas_course.rb
```diff
@@ -54,7 +65,72 @@ def self.from_metadata(metadata)
...
+    # config file
+    ##
+    def upload_scorm_package(scorm_package, course_id, tmp_name)
+      zip = scorm_package.write_zip tmp_name
+      RestClient.post(
+        "#{Blacklight.scorm_url}/api/scorm_courses",
+        {
+          oauth_consumer_key: "scorm-player",
+          lms_course_id: course_id,
+          file: File.new(zip, "rb"),
```
jamis, line 108: Referencing `File.new` this way opens a file, but never closes it. Better practice would be to call `File.open` with a block, and put the `RestClient.post` inside the block, like this:

```ruby
File.open(zip, "rb") do |file|
  RestClient.post(
      ...,
      file: file,
      ...,
  )
  # etc.
end
```

This will ensure that the file is automatically closed when the block terminates. (Ruby takes care of a lot of things for you, but managing IO still requires some attention, alas!)

lib/blacklight/models/file.rb
jamis: As this is a class variable, you can (and probably should) initialize it outside the method, rather than here. Using `||=` with a nil looks odd (though in this case, it does work), since what you're usually after with that idiom is to assign a non-nil value if the variable is currently nil (or false).

So try something like this:

```ruby
@@dir = nil
def self.cleanup
   # etc.
end
```

lib/blacklight/models/scorm_package.rb
```diff
@@ -0,0 +1,89 @@
...
+    # Returns true if a manifest is a scorm manifest file, false otherwise
+    ##
+    def self.scorm_manifest?(manifest)
+      parsed_manifest = Nokogiri::XML(manifest.get_input_stream.read)
+      schema_name = parsed_manifest.
+        xpath("//xmlns:metadata/xmlns:schema").
+        text.delete(" ").downcase
+      return schema_name == SCORM_SCHEMA
+    # NOTE we occasionally run into malformed manifest files
+    rescue Nokogiri::XML::XPath::SyntaxError
```
jamis, line 27: Would it be worth writing a warning to `STDERR` (or to a log file) here? Silently swallowing exceptions can often lead to _extremely_ frustrating debugging experiences.

jamis (commented):
I know I missed the window on reviewing this, but I figured I'd offer a few non-critical points for consideration. /cc @bfcoder

lib/blacklight/canvas_course.rb
jamis: I should add -- the file _is_ eventually closed, when the garbage collector gets around to noticing that the value is no longer referenced anywhere, but there's no guarantee of when that can happen. If you're opening files in a loop, for instance, you can easily encounter cases where you run out of file handles, because the GC isn't cleaning them up fast enough. It's just generally better to be explicit about it.

PR #15, opened 2016-12-15
Fix question item with no title [#136195417]

bfcoder:
This ensures that a title was found before grabbing the value from it.


kamijean (approved):
Looks good

PR #18, opened 2016-12-20
Fix several small things

bfcoder:
Write to STDERR when scorm manifest is malformed
Don’t interpolate the launch_url
Improve logging when uploading and creating scorm packages
Open file in a block to auto close it when posting to scorm server
Open file in a block to auto close it when posting to s3
Move get_scorm_packages to scorm_package.rb
Set `@@dir` to nil on the class instead of in a method
Open blackboard zip in a block. File will always be there from the rake tasks as they depend on it being there. So don’t check that the file exists.
Fix spec helper to accept any course in the stub
Fix specs to close fixtures

### DIFF
--- a/spec/models/scorm_package_spec.rb
+++ b/spec/models/scorm_package_spec.rb
@@ -12,32 +12,33 @@ end
 
 describe "ScormPackage" do
   it "should find all entries in same directory as manifest" do
-    zip, manifest = get_zip_manifest("scorm_package.zip")
-
-    result = Blacklight::ScormPackage.get_entries(zip, manifest)
-    assert_equal(result.size, 2)
-    assert_equal(
-      manifest.get_input_stream.read.include?("ADL SCORM"),
-      true,
-    )
+    get_zip_manifest("scorm_package.zip") do |zip, manifest|
+      result = Blacklight::ScormPackage.get_entries(zip, manifest)
+      assert_equal(result.size, 2)
+      assert_equal(
+        manifest.get_input_stream.read.include?("ADL SCORM"),
+        true,
+      )
+    end
   end
 
   it "should convert to zip file" do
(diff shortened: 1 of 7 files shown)


spec/models/scorm_package_spec.rb
```diff
@@ -12,8 +12,7 @@ def get_manifest_entry(zip)
 
 describe "ScormPackage" do
   it "should find all entries in same directory as manifest" do
-    zip = Zip::File.new("spec/fixtures/scorm_package.zip")
-    manifest = get_manifest_entry zip
+    zip, manifest = get_zip_manifest("scorm_package.zip")
```
jamis: The thing to be careful of, here, is that `zip` will have been closed by `get_zip_fixture`. I'm not 100% sure of the semantics surrounding `Zip::File.open`, but I suspect it works similarly to `File.open`, where the file is closed when the block terminates. If you need that zip file handle to remain open, you should yield `zip` and `manifest` to a block in turn, rather than returning them directly.

spec/models/scorm_package_spec.rb
```diff
@@ -24,13 +23,14 @@ def get_manifest_entry(zip)
   end
 
   it "should convert to zip file" do
-    zip = Zip::File.new("spec/fixtures/scorm_package.zip")
-    package = Blacklight::ScormPackage.new(zip, get_manifest_entry(zip))
+    zip, manifest = get_zip_manifest("scorm_package.zip")
```
jamis: See previous comment -- it would be better for `get_zip_manifest` to yield `zip` and `manifest` to a block, rather than returning them, in order to keep the `zip` handle open.

jamis (commented):
This looks really good! Just two issues with the specs, which I've noted.

PR #19, opened 2016-12-20
Handle staffinfo

nick-benoit14:
* Adds support for 'STAFFINFO' .dat files


lib/blacklight/xml_parser.rb
```diff
@@ -11,6 +11,22 @@ module Blacklight
     course: "Course",
     questestinterop: "Assessment",
     content: "Content",
+    staffinfo: "StaffInfo",
+
+    # categories: :iterate_categories,
```
bfcoder: Why add these back in? Do we still need them?

lib/blacklight/models/staff_info.rb
```diff
@@ -0,0 +1,76 @@
...
+        </ul>
+      HTML
+    end
+
+    def canvas_conversion(course)
+      page = CanvasCc::CanvasCC::Models::Page.new
+      page.body = construct_body
+      page.identifier = @id
+
+      page.page_name = if @title.empty?
```
bfcoder: With so little code maybe consider one line?
```ruby
page.page_name = @title.empty? ? "Contact" : @title
```

srphillips (approved):
I'm srphillips and I approve this merge request.

bfcoder (approved):
Looks good!

lib/blacklight/models/staff_info.rb
```diff
@@ -0,0 +1,71 @@
+
+module Blacklight
+  class StaffInfo
+    attr_accessor(
```
kamijean: Jamis was telling me that unless you need to edit the values on the class from another file then have this be attr_reader. In looking at the code, I don't know if you are accessing very many of these except in your specs. And then to access private variables you can do something like (@question.instance_variable_get :@max_score).

PR #20, opened 2016-12-21
Fix image/link src/href

bfcoder:
Add a base resource class that all models should inherit from.
Pass a collection of resources to each model.
Extract the correct image src from the resources.
Fix image src
Fix `<a>` tag href


conradbeach:
I tested the functionality of this branch on several Blackboard content types and fields. Embedded images were working everywhere I looked.

conradbeach:
One possible issue I noticed. If you go to the "Advanced" tab in the Blackboard "Insert/Edit Image" tool. There's a way to add alternative images to be displayed on mouseover and mouseout.

image

Blackboard stores that data by adding `onmouseover` and `onmouseout` attributes to the image tag that change the image src to the alt images. Blacklight doesn't fix those srcs like it does the primary src. Either way, the `onmouseover` and `onmouseout` attributes get removed by Canvas on import.

Do we want to try and handle these alternative images?

bfcoder:
@conradbeach I'm going to say no unless they specifically want that. I'm not sure that Canvas would even support such a thing.

lib/blacklight/models/resource.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def matches_xid?(_xid)
+      false
+    end
+
+    def _search_and_replace(resources, node_html, tag, attr)
+      node_html.search(tag).each do |element|
+        original_src = element[attr]
+        xid = original_src.split("/").last
+        file_resource = resources.detect_xid(xid)
```
nick-benoit14, line 24: Is there a way we could do this so it would work on files that don't have an xid associated with them? I think that is the case for at least some course files.

nick-benoit14: I just looked through a few courses and I couldn't find any instances of files being referenced with an href or a src attribute that did not have an xid, but there are courses that have files that do not include an xid. I am not sure however if they are ever linked to from other documents in the course. For what that is worth.

bfcoder: k, let's just assume the xid will be there then. If we run into a case where it isn't, we can address it then.

conradbeach: I also haven't seen an instance of a file being referenced without an xid, so I think this current implementation is a good choice for now.

lib/blacklight/models/module_item.rb
```diff
@@ -8,7 +10,7 @@ def initialize(title, type, identifierref)
       @workflow_state = "active"
     end
 
-    def canvas_conversion
+    def canvas_conversion(*)
```
nick-benoit14, line 13: I didn't know you could do this. That is real slick

bfcoder: yeah, same as doing `def meh(*args)` but in this case I didn't care about the variable as it isn't used.

lib/blacklight/models/file.rb
```diff
@@ -11,11 +13,17 @@ class BlacklightFile
     def initialize(zip_entry)
       path = zip_entry.name
       id = File.basename(path)
+      xid = id[/__(xid-[0-9]+_[0-9]+)/, 1]
```
nick-benoit14, line 16: I can't seem to find any documentation for matching strings this way. Do you know what it is called?

nick-benoit14: Also is it possible to assign that regex to a variable so it does not have to be duplicated?

conradbeach: I think [this](https://ruby-doc.org/core-2.2.0/String.html#method-i-5B-5D) is what you're looking for.

bfcoder: Where is it duplicated?

nick-benoit14: @conradbeach Yes, that is super slick. I didn't know you could do that.
@bfcoder The same regex is used on the next line to remove the xid from the name

bfcoder: It is close, but it is not the same.

nick-benoit14: @bfcoder Oh you're right. My mistake.

nick-benoit14:
Code looks good to me. I am having some issues getting course content to successfully upload, so I have not yet been to witness it fixing embedded images.

nick-benoit14:
I finally got it to work! And the images look great!

PR #21, opened 2016-12-21
Handle course files

nick-benoit14:
* Handles export files that are outside of csfiles/home_dir
* Moves file validity checking to file classes


spec/models/file_spec.rb
```diff
@@ -27,4 +27,52 @@
...
+
+    file_names = mock_entries.map(&:name).sort
+    result = mock_entries.map do |entry|
+      BlacklightFile.metadata_file?(file_names, entry)
+    end
+
+    expected_result = [false, true, false, true]
+
+    expected_result.each_with_index do |res, i|
+      assert_equal(res, result[i])
```
conradbeach: Is it helpful to compare each element of `result` and `expected_result` against each other? Would it be better just to compare both result arrays with `assert_equal(expected_result, result)`?
That's how the next spec does it.

conradbeach (commented):
Code looks good. I had just one suggested change that I noted.

As far as functionality goes, things look to be working correctly for the most part. I had an issue where the Blackboard package I was testing with had 2 `syllabus.xml` files in separate directories. Blacklight only put one of them in the .imscc, presumably overwriting the first. We should probably take a closer look at the possible ramifications of that.

spec/models/file_spec.rb
nick-benoit14: No you are correct, it is better to just assert_equal on the entire array. I forgot you could do that with minitest though. I fixed one of them, but forgot this one.

nick-benoit14:
You have a good point about file name collisions. This does not address file name collisions, but I don't think that the earlier code addresses them either. I think right now, before this pull request we were kind of flattening the file structure. Which definitely could be a problem. What you are talking about does need to be addressed though I think. 

I wonder if we should get James' branch merged, and then see what the best solution with the least code rewriting will be.

conradbeach:
Seems like a reasonable plan to me.

conradbeach (approved):
Changes look good to me.

PR #22, opened 2016-12-22
Add page url link to the body

bfcoder:
Some pages have a url in the title as a clickable link.
Canvas doesn’t support that, so just prepend that link to the body.

From blackboard:
image
`Step 2: Video clip from The Hurt Locker` is a link to vimeo.

Now in canvas:
image

### DIFF
--- a/lib/blacklight/models/wikipage.rb
+++ b/lib/blacklight/models/wikipage.rb
@@ -8,6 +8,14 @@ module Blacklight
           select { |p| p.title.start_with? @title }.count
         @title = "#{@title}-#{page_count + 1}" if page_count > 0
         page = CanvasCc::CanvasCC::Models::Page.new
+        if !@url.empty?
+          @body = %{
+            <a href="#{@url}">
+              #{@url}
+            </a>
+            #{@body}
+          }
+        end
         page.body = fix_html(@body, resources)
         page.identifier = @id
         page.page_name = @title
--- a/lib/blacklight/models/content.rb
+++ b/lib/blacklight/models/content.rb
@@ -36,6 +36,7 @@ module Blacklight
     def iterate_xml(xml, pre_data)
       @points = pre_data[:points] || 0
       @title = xml.xpath("/CONTENT/TITLE/@value").first.text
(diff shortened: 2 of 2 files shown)


lib/blacklight/models/wikipage.rb
```diff
@@ -8,6 +8,14 @@ def canvas_conversion(course, resources)
           select { |p| p.title.start_with? @title }.count
         @title = "#{@title}-#{page_count + 1}" if page_count > 0
         page = CanvasCc::CanvasCC::Models::Page.new
+        if !@url.empty?
+          @body = %{
```
nick-benoit14, line 12: Would it make sense to append the url to the end instead of replacing the body? In case there was a description that also had a url?

bfcoder: No, because the link should be at the top, not at the bottom of the content.

nick-benoit14: Won't this replace anything else that is in the body though?

bfcoder: Nope. See that I'm placing the body in there?

bfcoder: I have example images in the PR too that show it doesn't.

conradbeach:
These changes look good to me. Functions correctly from what I see.

lib/blacklight/models/wikipage.rb
nick-benoit14: I could be misunderstanding the flow somewhere, but I was thinking of this sort of timeline: 
```
iterate_xml is called on a Content object --> @body is set with some content if it exists --> canvas_conversion is called to create a wiki page
```
With this timeline if you call `@body = ` during canvas_conversion then if there was any sort of description being stored in @body it would be replaced. 

Where am I misunderstanding?

nick-benoit14: Oh just kidding I see how it is working.

nick-benoit14:
Looks good to me as well.

PR #23, opened 2016-12-22
excludes directories from course files

nick-benoit14:
* Don't include directories in course files

### DIFF
--- a/lib/blacklight/xml_parser.rb
+++ b/lib/blacklight/xml_parser.rb
@@ -113,7 +113,8 @@ module Blacklight
   # non-metadata file.
   ##
   def self.iterate_files(zipfile)
-    files = zipfile.glob("*/**/**")
+    files = zipfile.entries.select(&:file?)
+
     file_names = files.map(&:name)
     scorm_paths = ScormPackage.find_scorm_paths(zipfile)
 
--- a/spec/mocks/mockzip.rb
+++ b/spec/mocks/mockzip.rb
@@ -6,6 +6,10 @@ class MockZip
       @ftype = ftype
     end
 
+    def file?
+      true
+    end
+
     def extract(dummy = nil) end
   end
 


lib/blacklight/xml_parser.rb
```diff
@@ -113,7 +113,8 @@ def self.get_master_parent(pre_data, parents_ids, parent_id)
   # non-metadata file.
   ##
   def self.iterate_files(zipfile)
-    files = zipfile.glob("*/**/**")
+    files = zipfile.entries.select(&:file?)
```
conradbeach, line 116: I'm not sure this is doing what it's supposed to. When I tested it, the original implementation using `glob` didn't capture any directories. And this new implementation doesn't exclude anything new, it just captures some extra files (such as the .dat files).

bfcoder: I agree. If you were to change zipfile.glob to `zipfile.glob("**/**/**")` then it would catch everything including the files in the current directory. But I'm not sure that is what we want. As that includes the imsmanifest.xml and the .dat files. But for sure the glob isn't capturing the directories.

nick-benoit14: When I tried with `zipfile.glob("**/**/**")` I was getting directories included. In the documentation it looks like it just maps over all the entries in the zip, not differentiating between files and directories. 
http://www.rubydoc.info/github/rubyzip/rubyzip/Zip/EntrySet#glob-instance_method

So we could do `zipfile.glob('**/**/**').select(&:file?)` instead, but is that different than `zipfile.entries.select(&:file?)`?

I am grabbing all of the files in the root of the export to get the files in the resource directories that are at the root of the export. Currently we are going through our list of files and then deciding which ones to throw away, I.E. *.dat files and things.

nick-benoit14: And it could be that you decide not to include the files in those resource directories at the root of the export, but I think even then it would be necessary to follow a glob with a select method to make sure all the entries are files.

bfcoder:
```ruby
(byebug) zipfile.glob("**/**").select(&:directory?).count
0
(byebug) zipfile.glob("**/**").select(&:file?).count
381
(byebug) zipfile.glob("**/**").count
381
```
That is on a zip with several directories containing files. So I'm curious to know how you are getting directories with this command?

nick-benoit14: That is really interesting. When I put a byebug in the actual method and then run `rake blacklight:imscc` on the export titled ExportFile_SYS_196_20161028072839 I get: 
```
zipfile.glob("**/**").select(&:directory?).size
=> 91
zipfile.glob("**/**").select(&:file?).size
=> 1660
zipfile.entries.count
=> 1751
```

bfcoder: That is very strange. I wonder why you get directories and we don't..?

jamis: Some thoughts: @bfcoder -- have you tried it with the same export that @nick-benoit14 is using? Also, bundler should be ensuring you're all on the same version of ruby-zip, but it might be worth double-checking that. Lastly, you should make sure you're all doing this exactly the same -- are @bfcoder and @nick-benoit14 both checking `zipfile.glob` at the same point in the code, via `byebug`? A `zipfile.inspect` might be helpful, too (to help catch any unexpected differences, for instance).

jamis: Also, it might be worth mentioning that it could depend on how the zip file was created. Some tools might generate zip files where directories have separate entries (like @nick-benoit14 is seeing) while others might generate zip files where the directories are implied by the file paths (and so wouldn't have explicit directory entries).

nick-benoit14: I think that must have been it. I must have at some point rezipped the package that I was using. I downloaded a fresh zip and it did not include any directories. That is definitely something to keep in mind.

bfcoder: I get the same results on the zip that @nick-benoit14 was using.
In light of this, we should guard against this. Never know what we will get.
I'm just not sure which is best.
```ruby
files = zipfile.entries.select(&:file?)
```
or
```ruby
files = zipfile.glob("**/**").select(&:file?)
```

@jamis which do you recommend? glob calls `entries.map` with the pattern so we can get fancy if we need to.

I'm almost inclined to say we use `glob`, but I haven't convinced myself 100%. But I can see just using `entries` as it is less code. So maybe we stick with using entries?

jamis: If we're always selecting all files (via `**/**`) then we might as well just use `entries`. It's simpler, clearer, and easier to understand at a glance. If we ever need to match a different pattern, we can refactor that back to `glob`.

bfcoder (approved):
Looks good! Merge away!

PR #24, opened 2016-12-23
Fix fill in the blank answers

bfcoder:
Make sure an answer is found for fill in the blank. Surveys don't have an answer text, so none will be found.


conradbeach (approved):
This looks good to me.

PR #27, opened 2016-12-27
Fix requires to make gem work

bfcoder:
Remove require_all gem
Require each model individually
Remove byebug requires

This fixes the gem to work.


lib/senkyoshi/xml_parser.rb
```diff
@@ -1,6 +1,44 @@
-require "require_all"
-require_all "lib/senkyoshi/models/**/*.rb"
```
nick-benoit14, line 2: What is the motivation for moving away from require_all?

bfcoder: because it doesn't work with bundled gems. :]

bfcoder: It's also better practice to require each file you need in a gem. Rather than doing a wildcard.

nick-benoit14: Okie doke. Just curious. That sounds good though!

nick-benoit14 (approved):
Looks good to me. Nicely done!

PR #29, opened 2016-12-27
Filter Duplicate Scorm Packages

nick-benoit14:
* Handles duplicate scorm packages 
* Redesigns how scorm packages are located 
* Extracts scorm package points possible from blackboard export

This package previously would include the same scorm package twice --vvv--
image


lib/senkyoshi/models/scorm_package.rb
```diff
@@ -1,52 +1,65 @@
...
-      find_scorm_manifests(blackboard_export).map do |manifest|
-        ScormPackage.new blackboard_export, manifest
+    def self.find_scorm_item_paths(zip_file)
+      Nokogiri::XML.parse(
+        Senkyoshi.read_file(zip_file, "imsmanifest.xml"),
+      ).
+        xpath("//resource[@type='resource/x-plugin-scormengine']").
+        map { |r| r.xpath("./@bb:file").text }
+    rescue Exceptions::MissingFileError
+      []
```
bfcoder, line 41: You probably should be logging to STDERR instead of just swallowing the exception.

lib/senkyoshi/models/scorm_package.rb
```diff
@@ -1,52 +1,70 @@
...
     ##
-    # Extracts scorm packages from a blackboard export zip file
+    # Returns array of parsed scormItem files
     ##
-    def self.get_scorm_packages(blackboard_export)
-      find_scorm_manifests(blackboard_export).map do |manifest|
-        ScormPackage.new blackboard_export, manifest
+    def self.find_scorm_items(zip_file)
+      find_scorm_item_paths(zip_file).map do |path|
+        Nokogiri::XML.parse(zip_file.get_input_stream(path).read)
```
kamijean: You may want an exception here to catch this file if it isn't able to read it or just call the Senkyoshi read_file and let it handle it.

kamijean (approved):

PR #30, opened 2016-12-27
Matching

kamijean:
Fixing .text error from the testers bug find and adding distractors to the matching question


lib/senkyoshi/models/questions/matching.rb
```diff
@@ -27,17 +30,20 @@ def iterate_xml(data)
           flow_label = response.at("flow_label")
           flow_label.children.each_with_index do |label, index|
             if label.attributes["ident"].value == answer_id
-              answer = matches_array[index]
+              answer = Nokogiri::HTML(matches_array[index]).text
```
bfcoder: Answers can have HTML so you can't pull just the text out of it.

kamijean: For these types of questions it was still showing the p tags in the dropdown for some reason if you don't take them out.

conradbeach (approved):
The code and functionality looks good to me.

PR #31, opened 2016-12-27
Fixing when it breaks in the fill in the last blank question and when the resources don't have an element attribute

kamijean:


lib/senkyoshi/models/questions/fill_in_blank.rb
```diff
@@ -4,7 +4,9 @@ module Senkyoshi
   class FillInBlank < Question
     def iterate_xml(data)
       super
-      conditionvar = data.at("resprocessing").at("conditionvar")
+      if data.at("resprocessing")
+        conditionvar = data.at("resprocessing").at("conditionvar")
+      end
```
conradbeach, line 9: Should this same change be made in the `FillInBlankPlus` class?

bfcoder: It probably should. Nice catch. 👍

conradbeach (approved):
These changes look good to me.

PR #32, opened 2016-12-28
Quiz Instructions

conradbeach:
This changes the `Assessment` class so it adds both the Blackboard test/survey instructions and the test/survey description into the Canvas quiz instructions.


kamijean (approved):
Looks good

PR #33, opened 2016-12-29
Diversify Modules

nick-benoit14:
* Adds module item types like ExternalUrl and Attachment
* Improves module naming 
* Improves module grouping


lib/senkyoshi/models/attachment.rb
```diff
@@ -0,0 +1,22 @@
+require "senkyoshi/models/content"
+require "senkyoshi/models/module_item"
+require "byebug"
```
bfcoder: Remove byebug require

lib/senkyoshi/models/content_file.rb
```diff
@@ -6,8 +6,12 @@ class ContentFile < Resource
 
     def initialize(xml)
       @id = xml.xpath("./@id").first.text
-      @name = xml.xpath("./NAME").first.text
       @linkname = xml.xpath("./LINKNAME/@value").first.text
+      @name = xml.xpath("./NAME").first.text
+
+      # Remove leading slash if necessary so that ContentFile.name will match
+      # the Senkyoshi.xid
+      @name = @name[1, @name.length] if @name.start_with? "/"
```
bfcoder: use `[1..-1]` instead
```ruby
@name[1..-1]
```

lib/senkyoshi/xml_parser.rb
```diff
@@ -124,26 +127,30 @@ def self.connect_content(pre_data)
     pre_data["content"]
   end
 
-  def self.build_heirarchy(pre_data)
-    parents_ids = pre_data.
-      select { |p| p[:parent_id] == "{unset id}" }.
-      map { |u| u[:id] }
+  def self.build_heirarchy(organizations, pre_data)
+    unset_id = "{unset id}"
```
bfcoder, line 132: What does `"{unset id}"` even mean?

nick-benoit14: That is the value of `<ParentID>` for a top level folder I believe. 
Edit: I forgot markdown doesn't handle unquoted angle braces :/

kamijean: It's the id that blackboard gives to their top level folders

bfcoder: So weird. Blackboard never ceases to make me wonder...

lib/senkyoshi/models/content.rb
```diff
@@ -21,12 +21,19 @@ class Content < Resource
...
 
-    attr_accessor(:title, :body, :id, :files)
+    attr_accessor(:title, :body, :id, :files, :url)
     attr_reader(:extendeddata)
 
-    def self.from(xml, pre_data)
+    def self.from(xml, pre_data, resource_xids)
       type = xml.xpath("/CONTENT/CONTENTHANDLER/@value").first.text
       type.slice! "resource/"
+      xml.xpath("//FILES/FILE").map do |file|
```
nick-benoit14: I'm not quite sure what this is doing. Could you give a brief explanation?:

lib/senkyoshi/models/content.rb
```diff
@@ -45,20 +53,23 @@ def iterate_xml(xml, pre_data)
       end
       @type = xml.xpath("/CONTENT/RENDERTYPE/@value").first.text
       @parent_id = pre_data[:parent_id]
-      bb_type = xml.xpath("/CONTENT/CONTENTHANDLER/@value").first.text
-      bb_type.slice! "resource/"
-      @module_type = CONTENT_TYPES[bb_type]
+      @module_type = self.class.name.gsub("Senkyoshi::", "")
```
nick-benoit14: I wonder if it would be a bit more robust to use a hash map to map the class type to the module type. I think it would be a bit more verbose, but it might be a bit more clear. And then if you come up with a class that is not the same name as a canvas module type it won't break.

lib/senkyoshi/models/content.rb
kamijean: So some of the attachments weren't getting attached because the files they were referencing didn't exist in the files that were exported. So what this is doing is first I am getting the list of files that were exported from the code in the senkyoshi.rb. Then it's eventually passing this into this file as the resources xids, and then this is going through the files that are said to be in the content files and checking to see if they exist in the list of files. Then if they exist it does nothing and continues the same way, but if the file doesn't exist it turns the attachment into a document -- Wikipage. Then when it creates the wikipage it lets the user know that the file didn't exist with a note saying which files don't exist instead of not creating the attachment at all. Not sure if it's ideal, but if you have any suggestions let me know.

lib/senkyoshi/models/content.rb
kamijean: I am doing this because I am changing Attachment files to Wikipages if their files don't exist in the export. And so this is essentially just saying what is the class name minus the Senkyoshi:: part. 

The only one we are changing thus far is the quizzes to be "Quiz:Quizzes" because that can't be a class name. So maybe we would do it similarly to the way we are changing the quiz name to be the correct canvas module?

lib/senkyoshi/models/content.rb
bfcoder: Don't use map if you don't need to make an array. Use each instead.

lib/senkyoshi/models/content.rb
nick-benoit14: I was thinking about having something like this? 
```
    MODULE_TYPES = {
      "Senkyoshi::Attachment" => "Attachment",
      "Senkyoshi::Assignment" => "Assignment",
      "Senkyoshi::ExternalUrl" => "ExternalUrl",
      "Senkyoshi::WikiPage" => "WikiPage",
      "Senkyoshi::Quiz" => "Quizzes::Quiz",
    }.freeze
```
I added it to the code to try it out, but I haven't pushed it up. I can though if you think it is a way to go. I kind of like it because then you can also get rid of the 'Quizzes::Quiz' special case and treat all of the module item types the same way.

lib/senkyoshi/models/wikipage.rb
```diff
@@ -18,7 +18,9 @@ def canvas_conversion(course, resources)
 
         # Add page links to page body
         @files.each do |file|
-          if course.files.detect { |f| f.identifier == file.name }
+          if canvas_file = course.files.detect { |f| f.identifier == file.name }
```
nick-benoit14, line 21: What if we moved this to the ContentFile canvas conversion method? Then only code in that class would be modifying it's own attributes.

lib/senkyoshi/models/assessment.rb
```diff
@@ -43,22 +43,24 @@ def iterate_xml(data, pre_data)
...
+      course.assignments << assignment
+      assessment = setup_assessment(assessment, assignment, resources)
+      course.assessments << assessment
       course
     end
 
     def setup_assessment(assessment, assignment, resources)
       assessment.title = @title
       assessment.description = fix_html(@description, resources)
+      if @items.count == 0
```
conradbeach: Would `.zero?` be better here? I see that rule is disabled in our Rubocop config though, so I'm guessing there's a reason `== 0` is being used instead.

lib/senkyoshi/models/content_file.rb
```diff
@@ -2,15 +2,28 @@
...
     end
 
-    def canvas_conversion(*)
+    ##
+    # Remove leading slash if necessary so that ContentFile.name will match
+    # the Senkyoshi.xid
+    ##
+    def self.clean_xid(xid)
+      xid[1..-1] if xid.start_with? "/"
+    end
```
conradbeach, line 23: This method returns `nil` if `xid` doesn't start with a slash. It should be returning the original `xid` if there's no leading slash, yes?

lib/senkyoshi/models/content_file.rb
```diff
@@ -2,15 +2,28 @@
...
+
+    def self.correct_linkname(canvas_file)
+      canvas_file.file_path.split("/").last
+    end
+
+    def canvas_conversion(canvas_file = nil)
+      @linkname = ContentFile.correct_linkname(canvas_file) if !canvas_file.nil?
```
conradbeach: Could we instead do `if canvas_file` instead of `if !canvas_file.nil?`?

spec/models/content_file_spec.rb
```diff
@@ -4,27 +4,35 @@
...
+      file.canvas_conversion(mock_entry),
+      "ADV &amp;amp; DisAdv.pdf",
     )
   end
+
+  it "should strip only leading slash from name" do
+    xml = get_fixture_xml "file.xml"
+    file = Senkyoshi::ContentFile.new(xml.xpath("//FILE"))
+    assert_equal(file.name.start_with?("/"), false)
+  end
```
conradbeach, line 38: I think it'd be better to use `refute(file.name.start_with?("/"))` over `assert_equal`.

Also, the description of this test says "should strip _only_ leading slash from name", but the spec doesn't actually test that it only removes the leading slash. It just tests that the leading slash was removed. I would change this spec's description or update it so that it tests that other slashes are not removed.

lib/senkyoshi/models/content_file.rb
nick-benoit14: Good catch! Thanks!

spec/models/content_file_spec.rb
nick-benoit14: That is covered in another spec, but it would probably be more clear if it was tested here as well.

conradbeach (approved):
These changes look good to me.

PR #34, opened 2016-12-30
Add support for extended data in pages [#136807415]

bfcoder:
Parse xml in the extended data and add that data to the body

Before:
image

After:
image

Reference courses:
https://au.blackboard.com/webapps/blackboard/content/listContent.jsp?content_id=_2407700_1&course_id=_57815_1&navItem=content_lesson_plan&mode=quick

Instructure's erroneous conversion:
https://atomicjolt.instructure.com/courses/1426/pages/ip-324-readiness-assessment-tools


lib/senkyoshi/models/wikipage.rb
```diff
@@ -30,5 +38,31 @@ def canvas_conversion(course, resources)
 
       course
     end
+
+    def _extendeddata(extendeddata)
+      Nokogiri::XML(extendeddata).
+        search("LessonPlanComponent").
+        map do |node|
+          visible = true?(node.search("vislableToStudents").attr("value").to_s)
```
nick-benoit14: Another approach to this would be to use an xpath search. I think it looks a bit, but I suppose that is pretty subjective. `node.xpath("./vislableToStudents/@value").text`

nick-benoit14 (approved):
Looks good to me

lib/senkyoshi/models/wikipage.rb
kamijean: Is vislable their spelling mistake or ours?

lib/senkyoshi/models/wikipage.rb
```diff
@@ -16,6 +18,12 @@ def canvas_conversion(course, resources)
             #{@body}
           }
         end
+        if @extendeddata
+          @body = %{
+            #{@body}
+            #{_extendeddata(@extendeddata)}
+          }
+        end
```
kamijean: I am just wondering since we have 2 if statement to set the body if we should pull them out into their own function set_body function instead of making this function keep getting longer?

lib/senkyoshi/models/wikipage.rb
```diff
@@ -2,6 +2,8 @@
 
 module Senkyoshi
   class WikiPage < Content
+    include Senkyoshi
+
```
kamijean, line 6: Not sure if you need this include Senkyoshi in this file. None of the others models have it.

lib/senkyoshi/models/wikipage.rb
bfcoder: @kamijean It is blackboard's mistake. Nice huh? lol
@nick-benoit14 I prefer to avoid xpath as much as possible.

lib/senkyoshi/models/wikipage.rb
bfcoder: this is so I can gain access to the `#true?` method I put into the base module.

kamijean (approved):

lib/senkyoshi/models/wikipage.rb
```diff
@@ -30,5 +25,49 @@ def canvas_conversion(course, resources)
...
+        search("LessonPlanComponent").
+        map do |node|
+          _component_label(node) + node.search("componentValue/@value").to_s
+        end.
+        compact.
+        join(" ")
+    end
+
+    def _component_label(node)
+      visible = true?(node.search("vislableToStudents/@value").to_s)
```
bfcoder, line 59: Yup, `vislableToStudents` is how blackboard spells it...

PR #35, opened 2017-01-03
Fix contact

nick-benoit14:
* Fixes StaffInfo to handle multiple contacts 
Previously when there were multiple 'StaffInfo' files we were trying to generate a 'Contact' page for each one, this wasn't happening due to a naming conflict bug however. This pull request makes it so that we will generate a single 'Contact' page with contact info from all of the 'StaffInfo' pages.


kamijean:
Update the branch to master and it seems fine.

PR #36, opened 2017-01-05
Fix colliding uploads

nick-benoit14:
* Removes check to see if course exists before creating a new course

Previously there was a check to see if a course already existed with the same title, because titles aren't unique this resulted in missing courses when there were colliding titles.

These are different courses that have the same title: 
screen shot 2017-01-05 at 11 20 58 am


kamijean:
Looks good

PR #39, opened 2017-01-09
Preserves the Blackboard File Structure [#137113977]

conradbeach:
This changes Senkyoshi so it keeps the same file structure for course files as the Blackboard export instead of flattening everything into one directory.


bfcoder (approved):
Code looks good to me. Pulled down and tested and works as expected.

PR #41, opened 2017-01-10
Fix survey to be converted as a graded_survey

bfcoder:
This makes it show up in the assignments.
If it is set to `survey`, that translates to an "Ungraded Survey"
in Canvas, and thus doesn’t show up in the assignments.


kamijean (approved):
looks good

PR #43, opened 2017-01-11
Remove class variables for directories.

bfcoder:
This fixes being able to use `-m` with the rake task.


nick-benoit14 (approved):
Looks good to me. Nice work!

PR #44, opened 2017-01-11
Getting options for the quizzes

kamijean:


bfcoder (approved):
code looks good. I'd prefer to use `xml.at().text` instead of `xml.xpath().first.text` but that is just me.
Pulled down and works as expected.

PR #45, opened 2017-01-11
Fix generating contact pages

bfcoder:
Remove class variables.
Use only one instance of StaffInfo.

This fixes the contact page to contain contact info for only the given course. Also fixes so it works with `-m` in the task.

### DIFF
--- a/lib/senkyoshi/models/staff_info.rb
+++ b/lib/senkyoshi/models/staff_info.rb
@@ -1,4 +1,5 @@
 require "senkyoshi/models/resource"
+require "active_support/core_ext/string"
 
 module Senkyoshi
   class StaffInfo < Resource
@@ -58,6 +59,10 @@ module Senkyoshi
       body << str if var && !var.empty?
     end
 
+    def humanize(symbol)
+      symbol.to_s.humanize.titleize
+    end
+
     def construct_body(opts)
       body = "<div>"
       append_str body, "<img src=#{opts[:image]}/>", opts[:image]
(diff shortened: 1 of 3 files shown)


lib/senkyoshi/models/staff_info.rb
```diff
@@ -42,51 +29,68 @@ def parse_name(contact)
...
+      append_str(
+        body,
+        "<li>Office Address: #{opts[:office_address]}</li>",
+        opts[:office_address],
+      )
+      append_str(
+        body,
+        "<li>Home Page: #{opts[:home_page]}</li>",
+        opts[:home_page],
+      )
```
conradbeach: Lines 68-84 are doing the same thing for each option. Could we dry it up like this?

```ruby
[:email, :phone, :office_hours, :office_address, :home_page].each do |key|
  append_str body, "<li>#{sym_to_title(key)}: #{opts[key]}</li>", opts[key]
end
```

The `sym_to_title` method could look something like this.
```ruby
def sym_to_title(symbol)
  symbol.to_s.gsub("_", " ").split(/\W/).map(&:capitalize).join(" ")
end
```

I feel like this would make it easier to read as well.

conradbeach (commented):
I had one suggestion but the code and functionality looks good.

lib/senkyoshi/models/staff_info.rb
bfcoder: 👍  I like that, but your sym_to_title isn't very easy to read. So I'll use Activesupport instead.

conradbeach (approved):
Looks great.

lib/senkyoshi/models/staff_info.rb
conradbeach: Ah, I didn't realize we had ActiveSupport. That's definitely a better choice than using what I wrote.

PR #48, opened 2017-01-12
Fixing the quiz titles to return correctly with specs

kamijean:


lib/senkyoshi/models/question_bank.rb
```diff
@@ -28,7 +28,7 @@ def create_items(question_bank, resources)
       question_bank.questions = []
       questions.each do |item|
         question = item.canvas_conversion(question_bank, resources)
-        question.material = clean_up_material(question)
+        clean_up_material(question)
```
bfcoder: Consider this instead? Try to keep it pure?
```ruby
question.material = clean_up_material(question.material)
```

```ruby
TAGS = {
  "<p><span size=\"2\" style=\"font-size: small;\">.</span></p>" => "",
  "<p>.</p>" => "",
}.freeze
TAGS_RE = Regexp.union(TAGS.keys)

def clean_up_material(material)
  if material
    material.gsub!(TAGS_RE, TAGS)
    material.strip!
  end
  material
end
```

nick-benoit14: Won't `gsub!` and `strip!` both modify the original string though?

bfcoder: you have a good point.
```ruby
material = material.gsub(TAGS_RE, TAGS)
material = material.strip
```
would be better

bfcoder (approved):
Looks good!

PR #49, opened 2017-01-12
Fix parsing when no assessments are present. [#137495983]

bfcoder:
Check if there are assessments before processing them.


nick-benoit14 (approved):
Looks good, and appears to fix https://www.pivotaltracker.com/n/projects/1886533

PR #51, opened 2017-01-12
 Update Assignment Groups 

nick-benoit14:
* Creates assignments for user created gradebook entries that aren't linked to any other content
* Converts all assignment groups listed in blackboard gradebook

Below is an example of converted assignments and assignment groups. This mirrors the behavior of the Instructure converter. The below image was captured after converting: ExportFile_AU014_16BC.Summer.2016.CCC_MAFCLC003_1_20161028012650.zip

screen shot 2017-01-12 at 4 46 15 pm

### DIFF
--- a/lib/senkyoshi/models/assignment_group.rb
+++ b/lib/senkyoshi/models/assignment_group.rb
@@ -25,9 +25,13 @@ module Senkyoshi
       group.canvas_conversion
     end
 
-    def self.find_or_create(course, category)
-      assignment_group = course.assignment_groups.
+    def self.find_group(course, category)
+      course.assignment_groups.
         detect { |a| a.title == category }
+    end
+
+    def self.find_or_create(course, category)
+      assignment_group = find_group(course, category)
       unless assignment_group
         assignment_group = AssignmentGroup.create_assignment_group(category)
         course.assignment_groups << assignment_group
--- a/lib/senkyoshi/models/assignment_group.rb
+++ b/lib/senkyoshi/models/assignment_group.rb
@@ -32,7 +32,7 @@ module Senkyoshi
 
     def self.find_or_create(course, category)
       assignment_group = find_group(course, category)
-      unless assignment_group
+      if !assignment_group
         assignment_group = AssignmentGroup.create_assignment_group(category)
         course.assignment_groups << assignment_group
       end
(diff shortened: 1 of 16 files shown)


lib/senkyoshi/models/gradebook.rb
```diff
@@ -23,5 +39,29 @@ def get_categories(data)
...
+    def convert_categories(course)
+      @categories.each do |category|
+        if AssignmentGroup.find_group(course, category.last).nil?
+          course.assignment_groups <<
+            AssignmentGroup.create_assignment_group(category.last)
+        end
+      end
+    end
+
+    def canvas_conversion(course, _ = nil)
```
bfcoder: if the param is going to be used, don't use `_` to define it. Call it `resources`

lib/senkyoshi/models/outcome_definition.rb
```diff
@@ -0,0 +1,51 @@
...
+      @category = category
+    end
+
+    def iterate_xml(xml)
+      @content_id = xml.xpath("./CONTENTID/@value").text
+      @asidataid = xml.xpath("./ASIDATAID/@value").text
+      @id = xml.xpath("./@id").text
+      @title = xml.xpath("./TITLE/@value").text
+      @points_possible = xml.xpath("./POINTSPOSSIBLE/@value").text
+      @is_user_created = xml.xpath("./ISUSERCREATED/@value").text == "true"
```
bfcoder: You will need to add `include Senkyoshi` to this class, but there is a method that does this check for you.

```ruby
true? xml.xpath("./ISUSERCREATED/@value").text
```

lib/senkyoshi/models/assignment_group.rb
```diff
@@ -3,21 +3,40 @@
...
+    end
+
+    def self.find_group(course, category)
+      course.assignment_groups.
+        detect { |a| a.title == category }
+    end
+
+    def self.find_or_create(course, category)
+      assignment_group = find_group(course, category)
+      unless assignment_group
```
bfcoder: Consider using
```ruby
if !assignment_group
```
instead as it is easier to read what is going on.

bfcoder (approved):
Code looks good.
Functions as described.

PR #53, opened 2017-01-17
Fix Broken Links [#137061817]

conradbeach:
This fixes broken links in a few different content types such as assignments and announcements.

### DIFF
--- a/lib/senkyoshi.rb
+++ b/lib/senkyoshi.rb
@@ -56,7 +56,7 @@ module Senkyoshi
     course = CanvasCc::CanvasCC::Models::Course.new
     course.course_code = zip_name
     resources.each do |resource|
-      # Skips resources that are files.
+      # Skips resources that are directories.
       next if resource.respond_to?(:location) && !File.file?(resource.location)
 
       course = resource.canvas_conversion(course, resources)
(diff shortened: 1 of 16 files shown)


lib/senkyoshi.rb
```diff
@@ -55,6 +56,9 @@ def self.create_canvas_course(resources, zip_name)
     course = CanvasCc::CanvasCC::Models::Course.new
     course.course_code = zip_name
     resources.each do |resource|
+      # Skips resources that are files.
```
bfcoder: should this say `aren't` ?

lib/senkyoshi/models/file.rb
```diff
@@ -32,8 +32,13 @@ def extract_file(entry)
       name = "#{@dir}/#{entry.name}"
       path = File.dirname(name)
       FileUtils.mkdir_p path unless Dir.exist? path
-      entry.extract(name)
-      name
+
+      begin
```
bfcoder, line 36: I'm not a fan of open ended rescues like this. Something unrelated could fail and it will just get swallowed up in the rescue. Is there another way you can check for failures?

bfcoder:
What is the purpose of skipping directories?

lib/senkyoshi.rb
srphillips: Yes it should, it was changed but not pushed, we are getting a new version to review now

lib/senkyoshi/models/file.rb
srphillips: We are going over this now

PR #56, opened 2017-01-20
Add methods for Canvas to call

bfcoder:
This adds methods that canvas can call to do the converting.
Changes config so it can be configurable by more than the yml.
Add configuration specs

This enables [the canvas plugin](https://github.com/atomicjolt/senkyoshi_canvas_plugin/pull/1) to work


lib/senkyoshi/canvas_course.rb
```diff
@@ -63,9 +62,49 @@ def self.from_metadata(metadata, blackboard_export = nil)
...
+    # Assembles the launch url with the course_id
+    ##
+    def _scorm_launch_url(package_id)
+      "#{Senkyoshi.configuration.scorm_launch_url}?course_id=#{package_id}"
+    end
+
+    ##
+    # Creates a scorm assignment from a Canvas course object
+    ##
+    def _create_scorm_assignment_local(scorm_package)
```
nick-benoit14, line 83: When does this get used?

bfcoder: Line 67 :] If you look at https://github.com/atomicjolt/senkyoshi_canvas_plugin/pull/1/files#diff-c177a9387bec8249730ca0827fc439c5R151 it calls `process_scorm(local: true)` and that where this is used.

nick-benoit14: I saw line 67, I just couldn't figure out where in the world the 'if local' branch was getting run. That makes sense though how ha.

bfcoder: haha 👍 :)

nick-benoit14:
Appears to not be playing nice with production adhesion 1/23 8:18 AM. Out of 5 trials, one successfully launched. Others failed with one of: 
* 500 Internal server error
* 401 Unauthorized 
* 404 Not found

bfcoder:
I'm not sure how adhesion not working has to do with this PR though?

nick-benoit14:
I was just posting what I found from verifying that the pull request was working as expected. I thought it was pertinent that features that were touched with this pull request are not working correctly. It might be worth doing a quick check that adhesion is broken, and that some configuration hadn't been accidentally changed or something. That being said however I didn't notice anything that should have had an effect looking through the code.

nick-benoit14:
The same thing is happening on master though, so I guess we should just make sure that this behavior is explained by another bug that is already filed against adhesion.

nick-benoit14 (approved):
Changes look good. Courses are converted and upload as expected. Adhesion is returning several different errors both here and on master, so I was unable to verify that Scorm assignments still function as expected.

PR #57, opened 2017-01-20
Improve HTML in Matching Questions [#137407671]

conradbeach:
This fixes the issue we saw with answers in matching questions where the answers were cluttered with a bunch of raw HTML. This removes a selection of unhelpful HTML tags and leaves other potentially helpful ones. This cuts down on the clutter, but doesn't destroy potentially useful data. This also fixes broken links in matching question questions and answers.


lib/senkyoshi/models/questions/matching.rb
```diff
@@ -9,6 +9,16 @@ def initialize
       @distractors = []
     end
 
+    def strip_select_html(text)
+      text.gsub(/<\/?p[^>]*>/i, "").          # <p> tags
```
bfcoder: This is going to be uber slow. `gsub.gsub.gsub`...
Take a look at `models/question_bank.rb` at how I suggested the gsub be done there and mimic that. It will be much faster. https://github.com/atomicjolt/senkyoshi/blob/2eb3de8f214663f93e32cb71b63e45db4787a0d9/lib/senkyoshi/models/question_bank.rb
```ruby
TAGS = {
  "<p><span size=\"2\" style=\"font-size: small;\">.</span></p>" => "",
  "<p>.</p>" => "",
}.freeze
TAGS_RE = Regexp.union(TAGS.keys)
material = material.gsub(TAGS_RE, TAGS)
```

bfcoder: At a very minimum, it needs to be just one gsub

lib/senkyoshi/models/questions/matching.rb
```diff
@@ -9,6 +9,16 @@ def initialize
...
+        gsub(/<\/?b[^>]*>/i, "").             # <b> tags
+        gsub(/<\/?strong[^>]*>/i, "").        # <strong> tags
+        gsub(/<\/?em[^>]*>/i, "").            # <em> tags
+        gsub(/<\/?span[^>]*>/i, "").          # <span> tags
+        gsub(/<\/?i(?!mg)[^>]*>/i, "").       # <i> tags, ignores <img> tags
+        gsub(/ style=("|')[^"']*("|')/i, "")  # inline styles
+    end
```
conradbeach, line 24: Any suggestions on additions or deletions for this list of HTML tags would be welcome.

lib/senkyoshi/models/questions/matching.rb
conradbeach: Alright. I changed it to one `gsub`. How does that look?

lib/senkyoshi/models/questions/matching.rb
```diff
@@ -9,6 +9,18 @@ def initialize
       @distractors = []
     end
 
+    def strip_select_html(text)
+      tag_patterns = [/<\/?p[^>]*>/i,             # <p> tags
```
bfcoder: I would suggest making a constant TAG_PATTERNS defined outside of the method. Will be a bit faster that way. Same goes for the union. Make a constant TAGS_RE for them. Since it is always going to be the same data.

lib/senkyoshi/models/questions/matching.rb
conradbeach: I was curious so I did some (unscientific) testing and found that combining everything into one `gsub` looks to be 3 to 4 times faster.

lib/senkyoshi/models/questions/matching.rb
bfcoder: also, if you do it as a hash you can define what each pattern should be replaced with. Not that we need that exactly here, but just for your knowledge. :]

conradbeach: Ah, you're right. Those shouldn't be declared inside the method. I'll fix that.

lib/senkyoshi/models/questions/matching.rb
```diff
@@ -2,13 +2,25 @@
 
 module Senkyoshi
   class Matching < Question
+    TAGS_PATTERN = Regexp.union(/<\/?p[^>]*>/i,             # <p> tags
```
bfcoder: I'm not trying to beat a dead horse here...lol
if you were to multi line it would make it better to add patterns later.

```ruby
TAGS_PATTERN = Regexp.union(
  /<\/?p[^>]*>/i,             # <p> tags
  /<\/?b[^>]*>/i,             # <b> tags
  /<\/?strong[^>]*>/i,        # <strong> tags
  /<\/?em[^>]*>/i,            # <em> tags
  /<\/?span[^>]*>/i,          # <span> tags
  /<\/?i(?!mg)[^>]*>/i,       # <i> tags, no <img>
  / style=("|')[^"']*("|')/i, # inline styles
)
text.gsub(TAGS_PATTERN, "")
```
or
```ruby
TAG_PATTERNS = [
  /<\/?p[^>]*>/i,             # <p> tags
  /<\/?b[^>]*>/i,             # <b> tags
  /<\/?strong[^>]*>/i,        # <strong> tags
  /<\/?em[^>]*>/i,            # <em> tags
  /<\/?span[^>]*>/i,          # <span> tags
  /<\/?i(?!mg)[^>]*>/i,       # <i> tags, no <img>
  / style=("|')[^"']*("|')/i, # inline styles
].freeze

TAGS_RE = Regexp.union(TAG_PATTERNS)
text.gsub(TAGS_RE, "")
```
or
```ruby
TAG_PATTERNS = {
  /<\/?p[^>]*>/i => "",             # <p> tags
  /<\/?b[^>]*>/i => "",             # <b> tags
  /<\/?strong[^>]*>/i => "",        # <strong> tags
  /<\/?em[^>]*>/i => "",            # <em> tags
  /<\/?span[^>]*>/i => "",          # <span> tags
  /<\/?i(?!mg)[^>]*>/i => "",       # <i> tags, no <img>
  / style=("|')[^"']*("|')/i => "", # inline styles
}.freeze

TAGS_RE = Regexp.union(TAG_PATTERNS.keys)
text.gsub(TAGS_RE, TAG_PATTERNS)
```

conradbeach: No worries. Thanks for the suggestions.

bfcoder (approved):
Code looks good.

PR #58, opened 2017-01-23
Standardize Resource IDs

nick-benoit14:
* Adds RootResource class to represent 'resources' that are constructed from a single 'dat' file. 
* Changes all RootResources to have the 'dat' filename as the id

Blackboard gives each of the .'dat' files resources an 'id'. Unfortunately when these resources are referenced elsewhere, they are referenced by the file name instead of the id. This pull request standardizes the `@id` property on all 'RootResources', which are all 'Resources' that are created from a single `.dat` resource file. This makes it easier to find a resource by its filename (how I have often seen resources referenced) in a collection of resources. This is necessary for implementing module pre-requisites.


lib/senkyoshi/models/qti.rb
```diff
@@ -10,15 +10,16 @@
 }.freeze
 
 module Senkyoshi
-  class QTI < Resource
-    def self.from(data, pre_data)
+  class QTI < RootResource
+    def self.from(data, pre_data, _ = nil)
```
kamijean: You don't need the = nil, you can just do _

lib/senkyoshi/xml_parser.rb
```diff
@@ -81,25 +81,21 @@ def self.iterate_xml(organizations, resources, zip_file, resource_xids)
...
     end
   end
 
   def self.get_single_pre_data(pre_data, file)
-    pre_data.detect do |d|
+    found = pre_data.detect do |d|
       d[:file_name] == file || d[:assignment_id] == file
     end
+
+    found || { file_name: file }
```
kamijean: I wonder if you could just do

end || { file_name: file }

lib/senkyoshi/models/root_resource.rb
```diff
@@ -0,0 +1,23 @@
+require "senkyoshi/models/resource"
+
+module Senkyoshi
+  ##
+  # Class to represent a resource constructed from a single 'dat' file.
+  ##
+  class RootResource < Resource
```
kamijean: I wonder if the names should be switched because if the RootResource is inheriting from Resource and it feels like it should be the other way around.

nick-benoit14: Yeah I was having a real hard time coming up for a good name. Do you have any other ideas for a name? Something that captures that it is created from a single .dat file? I am not super happy with RootResource.

lib/senkyoshi/models/qti.rb
conradbeach: I think that depends on the use case though. If you want to be able to call `QTI.from` with either 2 or 3 arguments, you need the ` = nil`.

nick-benoit14: That is the case here. This is so that all of the root resources have the same signature, and we don't have to care which kind of root resource we are calling `from` on. Adding `_ = nil` just makes it so we won't have to go update every test that calls `from` to take `nil` as a third argument.

lib/senkyoshi/models/qti.rb
```diff
@@ -10,15 +10,16 @@
 }.freeze
 
 module Senkyoshi
-  class QTI < Resource
-    def self.from(data, pre_data)
+  class QTI < FileResource
+    def self.from(data, pre_data, _ = nil)
```
kamijean: You don't need the _ = nil -- just _
self.from(data, pre_data, _)

lib/senkyoshi/models/content.rb
```diff
@@ -30,12 +30,17 @@ class Content < Resource
       "Senkyoshi::Quiz" => "Quizzes::Quiz",
     }.freeze
 
-    attr_accessor(:title, :body, :id, :files, :url)
+    attr_accessor(:title, :body, :files, :url)
     attr_reader(:extendeddata)
 
+    def initialize(resource_id = nil)
+      super(resource_id)
+    end
```
conradbeach: What is the purpose of this `initialize` method? It's only calling `super`. Could we remove this method and just let it use the inherited `initialize` method?

lib/senkyoshi/models/file_resource.rb
```diff
@@ -0,0 +1,23 @@
...
+  ##
+  class FileResource < Resource
+    attr_reader(:id)
+
+    def initialize(id = nil)
+      @id = id
+    end
+
+    def self.from(xml = nil, pre_data = nil, _resource_xids = nil)
+      resource = new(pre_data[:file_name])
```
conradbeach, line 15: `pre_data[:file_name]` will throw a `NoMethodError` if `pre_data` is `nil`.

lib/senkyoshi/models/file_resource.rb
```diff
@@ -0,0 +1,23 @@
...
+      resource.iterate_xml(xml, pre_data)
```
conradbeach, line 16: Is it helpful to pass `xml` and `pre_data` to this method call when the method just ignores them?

conradbeach (changes requested):
A couple of possible changes.

lib/senkyoshi/models/file_resource.rb
nick-benoit14: Yeah it makes sense to remove `xml = nil`, `pre_data = nil`. I will do that. I'm not sure why I added that in the first place.

lib/senkyoshi/models/file_resource.rb
nick-benoit14: I'm not sure I'm following you. Isn't the method using both of those parameters?

lib/senkyoshi/models/content.rb
nick-benoit14: I didn't realize that it would automatically inherit the non-default constructor. It looks like it does though, and it can go away.

lib/senkyoshi/models/file_resource.rb
conradbeach: Perhaps I'm not reading the code correctly. It looks to me that we're creating a new instance of `FileResource` on line 15 and then calling `iterate_xml` on that instance. 
`FileResource`'s implementation of `iterate_xml` is 
```ruby
def iterate_xml(_xml, _pre_data)
  self
end
```
Neither of the arguments passed to it are used. Is that correct?

nick-benoit14: Oh my mistake, I thought you were talking about the above method. Each specialization of the FileResource implements some version of `iterate_xml`, so I guess I was using it as a way to communicate that everything that inherits from this method should implement this method to do something more than just return itself. It could probably go though, as it doesn't really do much of anything

conradbeach: Ok. Whatever you think makes the most sense.

conradbeach (approved):

lib/senkyoshi/models/qti.rb
bfcoder: I don't have a problem with doing `_ = nil` but I prefer to give it a name rather than just `_`. But that is just my ocd. :]

conradbeach: Names are good...

nick-benoit14: Agreed :)

PR #59, opened 2017-01-23
HTML Tags for Matching Question Answers [#137407671]

conradbeach:
Adds `<font>` tags to the list of HTML tags that get removed from matching question answers.


bfcoder (approved):
Looks good

PR #61, opened 2017-01-24
Module reorganized

kamijean:
Making it so the organizations of the blackboard folders match up with the modules in Canvas. Adding indentations for the folder structures and such.


lib/senkyoshi/models/attachment.rb
```diff
@@ -10,11 +10,13 @@ def iterate_xml(xml, pre_data)
...
         @files.first.name,
         @url,
+        @indent,
+        @file_name,
       ).canvas_conversion
       self
     end
 
-    def canvas_conversion(course, _resource)
+    def canvas_conversion(course, _)
```
bfcoder: Let's avoid naming just `_` and keep it with a name so it is clear what would have been passed in.

lib/senkyoshi/models/external_url.rb
```diff
@@ -2,7 +2,7 @@
 
 module Senkyoshi
   class ExternalUrl < Content
-    def canvas_conversion(course, _resource)
+    def canvas_conversion(course, _)
```
bfcoder: Again, keep it `_resource`

lib/senkyoshi/models/module_converter.rb
```diff
@@ -0,0 +1,87 @@
...
+    def self.add_canvas_module(course, data)
+      canvas_module = Module.new(data[:title], data[:file_name])
+      course.canvas_modules << canvas_module.canvas_conversion
+      course
+    end
+
+    def self.add_canvas_module_item(course, module_item, data)
+      parent_module = get_subheader_parent(course, data)
+      if !parent_module
+        parent_module = Module.new(course.title, "aj_main_module").
```
bfcoder: Make a constant in the senkyoshi.rb under `DIR_BASE` that defines "aj_main_module" so it can be re-used.

lib/senkyoshi/models/module_converter.rb
```diff
@@ -0,0 +1,87 @@
...
+    def self.get_subheader_parent(course, data)
+      if !data[:parent_id]
+        parent_module = course.canvas_modules.
+          detect { |a| a.identifier == "aj_main_module" }
+      else
+        parent_module = course.canvas_modules.
+          detect { |a| a.identifier == data[:parent_id] }
+        if !parent_module
+          course.canvas_modules.
+            reject { |a| a.title == "master_module" }.
```
bfcoder: Same as "aj_main_module", define one for "master_module"

lib/senkyoshi/models/quiz.rb
```diff
@@ -2,7 +2,7 @@
 
 module Senkyoshi
   class Quiz < Content
-    def canvas_conversion(course, _resource)
+    def canvas_conversion(course, _)
```
bfcoder: revert back to `_resource`

lib/senkyoshi.rb
```diff
@@ -41,9 +42,16 @@ def self.parse(zip_path, imscc_path)
...
         select { |r| r.include?("xid-") }
-      resources.add(Senkyoshi.parse_manifest(file, manifest, resource_xids))
 
-      course = create_canvas_course(resources, zip_path)
+      xml = Nokogiri::XML.parse(manifest)
+      xml_resources = xml.at("resources")
+      xml_organizations = xml.at("organizations")
+
+      pre_data = Senkyoshi.pre_iterator(xml_organizations, xml_resources, file)
+      resources.add(Senkyoshi.
```
bfcoder: Consider?
```ruby
resources.add(
  Senkyoshi.iterate_xml(xml_resources, file, resource_xids, pre_data),
)
```

lib/senkyoshi/models/content.rb
```diff
@@ -78,18 +80,14 @@ def iterate_xml(xml, pre_data)
...
-
-    def self.get_pre_data(xml, file_name)
-      id = xml.xpath("/CONTENT/@id").first.text
-      parent_id = xml.xpath("/CONTENT/PARENTID/@value").first.text
-      {
-        id: id,
-        parent_id: parent_id,
-        file_name: file_name,
-      }
+      ModuleItem.new(
```
bfcoder, line 83: Nice refactor 👍

lib/senkyoshi/models/content.rb
```diff
@@ -99,12 +97,11 @@ def canvas_conversion(course, _resources = nil)
     def create_module(course)
       course.canvas_modules ||= []
       cc_module = course.canvas_modules.
-        detect { |a| a.identifier == @parent_id }
+        detect { |a| a.title == "master_module" }
```
bfcoder: Use the const I mention later in the PR

lib/senkyoshi/models/content.rb
```diff
@@ -99,12 +97,11 @@ def canvas_conversion(course, _resources = nil)
...
       if cc_module
         cc_module.module_items << @module_item
       else
-        title = @parent_title || @title
-        cc_module = Module.new(title, @parent_id)
+        cc_module = Module.new("master_module", "master_module")
```
bfcoder: Would it make sense to have a new module default to these values?

lib/senkyoshi/models/heirarchy.rb
```diff
@@ -0,0 +1,82 @@
+require "senkyoshi/models/resource"
+require "byebug"
```
bfcoder: ack! :]

lib/senkyoshi/models/forum.rb
```diff
@@ -22,6 +32,18 @@ def canvas_conversion(course, _resources = nil)
       discussion.identifier = @id
       discussion.discussion_type = @discussion_type
       course.discussions << discussion
+      if @module_item
+        master_module = course.canvas_modules.
```
bfcoder: I'm seeing this same detect repeated. Consider making a method in the course to dry this up? Perhaps make a instance variable to hold the `canvas_modules` in course.rb?

lib/senkyoshi/models/file_resource.rb
```diff
@@ -19,5 +19,20 @@ def self.from(xml, pre_data, _resource_xids = nil)
     def iterate_xml(_xml, _pre_data)
       self
     end
+
+    def create_module(course, module_item)
+      course.canvas_modules ||= []
+      cc_module = course.canvas_modules.
```
bfcoder: This detect is still repeated in the code. My thoughts are to make an instance variable on a course that holds the canvas_modules. Make an `attr_accessible` for it. In the initializer for a course set it to an empty array. And write a method
```ruby
def master_module
  @canvas_modules.detect { |a| a.title == MASTER_MODULE }
end
```

now you can say `cc_module = course.master_module` here and elsewhere.

.gitignore
```diff
@@ -22,3 +22,4 @@
 .idea/
 senkyoshi.iml
 senkyoshi.yml
+npm-debug.log
```
bfcoder: add a newline to end of file

lib/senkyoshi/models/heirarchy.rb
```diff
@@ -0,0 +1,81 @@
...
+      get_indent(item.parent, indent)
+    end
+
+    def self.set_discussion_boards(discussion_boards, toc_item)
+      if toc_item[:internal_handle] == "discussion_board_entry"
+        resource = discussion_boards.select do |db|
+          title_attribute = db.attributes["title"] || db.attributes["bb:title"]
+          title_attribute.value == toc_item[:title]
+        end
+        if resource.count == 1
```
bfcoder, line 32: What happens if it is more than 1?

lib/senkyoshi/models/heirarchy.rb
```diff
@@ -0,0 +1,81 @@
...
+          file_attribute = resource[0].attributes["file"] ||
```
bfcoder: consider `resource.first` or
```ruby
FILE = 0
resource[FILE]
```
and make that number less magical.

lib/senkyoshi/models/heirarchy.rb
```diff
@@ -0,0 +1,81 @@
...
+          detect { |ct| ct[:original_file] == file_name } || {}
+        toc_item[:file_name] = file_name
+        toc_item[:title] = title
+        toc_item[:parent_id] = get_parent_id(course_toc, item_id)
+        toc_item
+      end
+    end
+
+    def self.get_parent_id(course_toc, item_id)
+      header_ids = course_toc.select { |ct| ct[:target_type] == "SUBHEADER" }.
```
bfcoder: Consider moving this select and map into a method to be reused and DRY this up.

lib/senkyoshi/models/heirarchy.rb
kamijean: Then it does nothing because you can't guarantee that it will connect the right discussion to the right location.

lib/senkyoshi/models/heirarchy.rb
```diff
@@ -0,0 +1,81 @@
...
+        map { |sh| sh[:original_file].gsub("res", "") }
+      if header_ids.empty?
+        header_ids = course_toc.select { |ct| ct[:target_type] == "CONTENT" }.
+          map { |sh| sh[:original_file].gsub("res", "") }
+      end
+      header_id = header_ids.
+        reject { |x| x.to_i > item_id.to_i }.
+        min_by { |x| (x.to_i - item_id.to_i).abs }
+
+      header_id ? "res" + header_id : nil
```
bfcoder: Consider interpolation as it is suppose to be faster.
```ruby
header_id ? "res #{header_id}" : nil
```

lib/senkyoshi/models/file_resource.rb
```diff
@@ -19,5 +19,20 @@ def self.from(xml, pre_data, _resource_xids = nil)
     def iterate_xml(_xml, _pre_data)
       self
     end
+
+    def create_module(course)
+      course.canvas_modules ||= []
+      cc_module = course.canvas_modules.
+        detect { |a| a.title == MASTER_MODULE }
```
bfcoder: This detect is still repeated in the code. My thoughts are to make an instance variable on a course that holds the canvas_modules. Make an `attr_accessible` for it. In the initializer for a course set it to an empty array. And write a method
```ruby
def master_module
  @canvas_modules.detect { |a| a.title == MASTER_MODULE }
end
```

now you can say `cc_module = course.master_module` here and elsewhere.

lib/senkyoshi/models/file_resource.rb
```diff
@@ -19,5 +19,19 @@ def self.from(xml, pre_data, _resource_xids = nil)
     def iterate_xml(_xml, _pre_data)
       self
     end
+
+    def create_module(course)
+      course.canvas_modules ||= []
+      cc_module = Course.master_module(course)
```
bfcoder, line 25: Why not have the course manage the modules? You wouldn't have to do the line above `course.canvas_modules ||= []` as `course` would already be aware of its `canvas_modules`

I mean you have dryed it up with the Class method. But I think an instance method would be better. Not a requirement, but my opinion. :]

kamijean: So course is a canvas_cc model and not ours, so I guess it makes things a little more complicated to do it that way since I don't have access to our course model to add an instance method there. I have been trying to think of a way. If you have insight let me know.

bfcoder: oh, I was thinking it was from course.rb. In that case we can't really do anything there. :/ bummer

kamijean: Yeah, I tried to do it a different way, but I just couldn't think of a better way to make it all flow better. Too many files and too many tools to figure out how to deal with them cleanly. But I am totally open to ideas.

lib/senkyoshi/models/file_resource.rb
```diff
@@ -19,5 +19,19 @@ def self.from(xml, pre_data, _resource_xids = nil)
...
+      if cc_module
+        cc_module.module_items << @module_item
+      else
+        cc_module = Module.new(MASTER_MODULE, MASTER_MODULE)
```
bfcoder, line 29: What would you think about making a new module default values to `MASTER_MODULE`? So you can just say `cc_module = Module.new`?

bfcoder: and if you do pass in values, then they will get assigned as normal.

bfcoder (approved):
Code looks good. Pulled down and it does create indented modules as expected.

lib/senkyoshi/models/file_resource.rb
kamijean: This seems like a good idea. Let me think about how to make that process work better like that and then I'll do a new branch to clean it up.

PR #64, opened 2017-01-25
Minor README clean up. Fix Rubocop offense.

conradbeach:


README.md
```diff
@@ -96,5 +96,5 @@ Bug reports and pull requests are welcome on GitHub at https://github.com/atomic
 The gem is available as open source under the terms of the [MIT License](http://opensource.org/licenses/MIT).
 
 
-### Things Not Quite Implemented
-People, Groups, Sets, Blogs
+## Things Not Quite Implemented
+People, Sets
```
conradbeach: What are people and sets? Is it accurate that they haven't been implemented?

kamijean: I think you can take off people because I am guessing that is referring to Staff, and I wonder if sets is possibly referring to rubric sets?

conradbeach: Ok. I'll remove people from the list.

I also wondered if sets was referring to question sets / question banks.

kamijean: That's probably what they are referring to actually. I think you could be right. I think it would be fine to remove it because we can't remember what exactly it's referring to. And then we can add things as we find them again.

conradbeach: Sounds good.

bfcoder (approved):
Looks good.

PR #65, opened 2017-01-26
Question fixes

kamijean:
Fixing the true/false question answers to compare by strings. And making the id of the matching questions to match our hex generator to fix the error.

### DIFF
--- a/lib/senkyoshi/models/question.rb
+++ b/lib/senkyoshi/models/question.rb
@@ -125,6 +125,7 @@ module Senkyoshi
         if score_number > 0
           @correct_answers["fraction"] = score_number.to_f / @max_score.to_f
         else
+          # mark as correct when there is no score for the answer
           @correct_answers["fraction"] = 1
         end
       end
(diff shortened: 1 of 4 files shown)


lib/senkyoshi/models/question.rb
```diff
@@ -125,7 +125,7 @@ def set_correct_answers(resprocessing)
         if score_number > 0
           @correct_answers["fraction"] = score_number.to_f / @max_score.to_f
         else
-          @correct_answers["fraction"] = 0
+          @correct_answers["fraction"] = 1
```
bfcoder, line 129: Can you explain what the significance of the `1` is?

bfcoder: Maybe add a comment explaining why it needs to be set to `1`?

kamijean: Blackboard allows for the questions to not be worth points when the question is contained in a survey. So this if else tells canvas that even if a problem isn't worth points it's still correct by marking it a 1. The 1 to canvas is correct and 0 is incorrect. Then any number in between is a percentage of the problem they will get correct if they answer it correctly which applies to like matching questions in which the student needs to get all the matching correct to get the full points. I'll make a comment for this too.

bfcoder (approved):
Looks good!

PR #66, opened 2017-01-27
Add Partial Support for Course Links

conradbeach:
This adds partial support for the course link content type from Blackboard. Senkyoshi turns the Blackboard course link into a Canvas wiki page and embeds a link in the page body. The link is not valid, but it provides enough information for the course editor to find the resource it should link to and fix it manually.

### DIFF
--- a/lib/senkyoshi/models/wikipage.rb
+++ b/lib/senkyoshi/models/wikipage.rb
@@ -35,10 +35,19 @@ module Senkyoshi
 
     def _set_body(original_body, url, extendeddata)
       body = original_body.dup
+
       if !url.empty?
         body = %{
           <a href="#{url}">
             #{url}
+          </a>
+          #{body}
+        }
+      end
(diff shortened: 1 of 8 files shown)


lib/senkyoshi/models/wikipage.rb
```diff
@@ -43,12 +44,21 @@ def _set_body(original_body, url, extendeddata)
           #{body}
         }
       end
+      if @referred_to_title.present?
+        body = %{
+          <a href="$CANVAS_COURSE_REFERENCE$#{@referred_to_title}">
+            Course Link: #{@referred_to_title}
```
conradbeach, line 50: I added this "Course Link: " label before the title to provide more context to users as to what it's supposed to be linking to. If you think it'd be better to leave this off or change it to something else, let me know.

kamijean (approved):

PR #67, opened 2017-01-30
Add 'Adaptive Release' support

nick-benoit14:
Blackboard exports have a concept of 'Adaptive Release'. These manifest themselves as `<Rule` .dat files. This pull request adds support for converting 'Rules', into module prerequisites, and completion requirements. Canvas does not support 'Adaptive Release' on a single item like blackboard does, so whenever we find a rule that makes a module item depend on another item in it's same module, it becomes a module completion requirement. When a module item depends on an item that is in a different module, that other module becomes a prerequisite for the current module in question.

screen shot 2017-01-29 at 2 16 35 pm

The above image shows an example of a module that has both completion requirements and prerequisites defined.

Here is a good example of a blackboard course that makes use of adaptive release https://au.blackboard.com/webapps/blackboard/execute/modulepage/view?course_id=_63520_1&cmp_tab_id=_137304_1&editMode=true&mode=cpview

* Adds `Rule` file resource
* Adds `RuleCriteria` class to handle different sorts of adaptive release criteria


lib/senkyoshi.rb
```diff
@@ -103,7 +111,7 @@ def self.initialize_course(canvas_file_path, blackboard_file_path)
     end
   end
 
-  def true?(obj)
+  def self.true?(obj)
```
bfcoder, line 114: This change is unnecessary. In the files you are using this, add to the class you are in `include Senkyoshi` and you will have access to this helper. See `outcome_definition.rb:6`

nick-benoit14:
I used a ruby 2.3 specific feature, this broke the tests for ruby versions < 2.3. Are we aiming to support ruby >= 2.3, or also lower version numbers? I was under the impression we were targeting 2.3.1.

bfcoder:
Originally we were trying to support 2.0, 2.1, 2.2, and 2.3. But now that we are moving to ruby 2.3.1 exclusively, it might be wise to drop support for the first three and go with just 2.3.1 and above. But that limits others that might not be using 2.3 from using this gem. The feature you are using is nice but I'm not sure it should keep others from using this gem when you can just as easily do an if statement to replace that feature.

nick-benoit14:
It would take about a minute to get rid of it, I just wasn't aware we were supporting lower versions. I will make that change.

lib/senkyoshi.rb
nick-benoit14: I made this change because I was wanted to use this method in a class method. I couldn't figure out a way to use a module instance method in a class method. 

I also noticed that this was the only non static method in this file. Is there something about this method in particular that makes it more suited to be an instance method?

lib/senkyoshi/collection.rb
```diff
@@ -16,6 +16,24 @@ def detect_xid(xid)
       end
     end
 
+    def find_by_id(id)
+      @resources.detect do |item|
```
bfcoder: I'm tempted to make this a one liner `@resources.detect { |item| item.respond_to?(:id) && item.id == id }` but that is up to you.

lib/senkyoshi.rb
bfcoder:
```ruby
module Foo
  def howdy?
    "hi"
  end
end

module Foo
  class Bar
    include Foo # for instance methods
    extend Foo # for class methods

    def meclass
      howdy?
    end

    def self.meinstance
      howdy?
    end
  end
end
```

Just wasn't a need to have it a class method. I didn't realize you were using it in class methods. So if you feel this is cleaner then that is fine with me.

lib/senkyoshi/models/grade_range_percent_criteria.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def get_points_possible(resources, id)
+      resource = resources.find_by_id(id)
+      resource.points_possible if resource
+    end
+
+    def get_percentage(min_score, points_possible)
+      possible = points_possible.to_f
+      (min_score.to_f / possible) * possible
```
bfcoder: Are you sure this is returning the percentage? It looks to me like it is returning the min_score that was passed in as a float.

nick-benoit14: You are correct. That was dumb of me.

bfcoder: You should probably write some tests for these methods. :]

lib/senkyoshi/models/outcome_definition.rb
```diff
@@ -10,7 +10,10 @@ def self.from(xml, category)
       outcome_definition.iterate_xml(xml)
     end
 
-    def initialize(category)
+    def initialize(category, id = nil, content_id = nil, asidataid = nil)
+      @id = id
```
bfcoder, line 14: should this be `id || Senkyoshi.create_random_hex` ?

lib/senkyoshi/models/rule_criteria.rb
```diff
@@ -0,0 +1,146 @@
...
+      new(RuleCriteria.get_id(xml), RuleCriteria.get_negated(xml))
+    end
+
+    ##
+    # Determine two identifierrefs are pointed to by items in the same module
+    def self.in_same_module?(modules, content_id, resource_id)
+      content_module = Module.find_module_from_item_id(modules, content_id)
+      resource_module = Module.find_module_from_item_id(modules, resource_id)
+      return nil if content_module.nil? || resource_module.nil?
+      !content_module.nil? && content_module == resource_module
```
srphillips: Change to remove extra nil check

lib/senkyoshi/models/outcome_definition.rb
```diff
@@ -10,7 +10,10 @@ def self.from(xml, category)
       outcome_definition.iterate_xml(xml)
     end
 
-    def initialize(category)
+    def initialize(category, id = nil, content_id = nil, asidataid = nil)
+      @id = id || Senkyoshi.create_random_hex
```
srphillips: Implement dummy id's on the specs instead of creating one here.  Real data won't be missing an id

lib/senkyoshi/models/questions/ordering.rb
```diff
@@ -32,9 +32,11 @@ def canvas_conversion(assessment, _resources = nil)
...
-        id = varequal.text
-        order_answers[id] = index + 1
+      if correct
+        correct.at("and").children.each_with_index do |varequal, index|
+          id = varequal.text
+          order_answers[id] = index + 1
+        end
       end
       order_answers
     end
```
srphillips, line 42: Review with @kamijean

lib/senkyoshi/models/rule.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def initialize(resource_id)
+      super(resource_id)
+      @criteria_list = []
+    end
+
+    def get_criteria_list(xml)
+      xml.children.select { |child_xml| !child_xml.blank? }.
+        map do |child_xml|
+          CRITERIA_MAP[child_xml.name.downcase.to_sym]&.from_xml(child_xml)
```
srphillips: This is awesome Ruby 2.3 syntax, but we need to support down to ruby 2.1.  Please fix to accommodate.

srphillips (approved):
I approve this merge

PR #69, opened 2017-02-01
Add Large File Support

nick-benoit14:
* Disables timeout for *.imscc upload 
* Adds support for 64 bit zip

### DIFF
--- a/lib/senkyoshi.rb
+++ b/lib/senkyoshi.rb
@@ -11,6 +11,8 @@ require "ostruct"
 require "nokogiri"
 require "zip"
 
+Zip.write_zip64_support = true
+
 require "senkyoshi/exceptions"
 
 module Senkyoshi
--- a/lib/senkyoshi.rb
+++ b/lib/senkyoshi.rb
@@ -96,7 +96,6 @@ module Senkyoshi
     end
 
     course = ModuleConverter.set_modules(course, pre_data)
-
     resources.find_instances_of(Rule).each do |rule|
       course = rule.canvas_conversion(course, resources)
     end
--- a/lib/senkyoshi/configuration.rb
+++ b/lib/senkyoshi/configuration.rb
@@ -17,7 +17,9 @@ module Senkyoshi
       @scorm_url = Configuration._config[:scorm_url]
       @scorm_launch_url = Configuration._config[:scorm_launch_url]
       @scorm_shared_auth = Configuration._config[:scorm_shared_auth]
-      @request_timeout = Configuration._config[:request_timeout]
+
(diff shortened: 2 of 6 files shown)


lib/senkyoshi.rb
```diff
@@ -11,6 +11,8 @@
 require "nokogiri"
 require "zip"
 
+Zip.write_zip64_support = true
```
bfcoder, line 14: Just my opinion, but I feel this is not in the scope of this PR and should be its own PR.

nick-benoit14: What if we changed the name of the PR to 'Add Large File Support'? Then I think it better reflects the reason we are changing these things.

bfcoder: Then it would be in scope. :] 👍

srphillips (approved):

bfcoder:
I would suggest we make the timeout configurable with the yml file. Have it default to 30 mins perhaps?

srphillips:
@bfcoder makes an excellent suggestion to put it into the config

lib/senkyoshi/configuration.rb
```diff
@@ -16,6 +17,7 @@ def initialize
       @scorm_url = Configuration._config[:scorm_url]
       @scorm_launch_url = Configuration._config[:scorm_launch_url]
       @scorm_shared_auth = Configuration._config[:scorm_shared_auth]
+      @request_timeout = Configuration._config[:request_timeout]
```
bfcoder:
```ruby
DEFAULT_TIMEOUT = 1_800 # 30 minutes
```
```ruby
@request_timeout = Configuration._config[:request_timeout] || DEFAULT_TIMEOUT
```

lib/senkyoshi/configuration.rb
```diff
@@ -16,6 +17,9 @@ def initialize
       @scorm_url = Configuration._config[:scorm_url]
       @scorm_launch_url = Configuration._config[:scorm_launch_url]
       @scorm_shared_auth = Configuration._config[:scorm_shared_auth]
+
+      # Default timeout is 30 minutes
+      @request_timeout = Configuration._config[:request_timeout] || 1800
```
bfcoder:
```ruby
TIMEOUT = 1_800 # 30 minutes
```
```ruby
@request_timeout = Configuration._config[:request_timeout] || TIMEOUT
```

bfcoder (approved):
Looks great!

bfcoder:
wait, still needs the readme to be updated

bfcoder (approved):
excellent

PR #72, opened 2017-02-03
Gracefully Handle Scorm Errors

nick-benoit14:
* Gracefully handle error messages from Scorm manager


srphillips (commented):
I'm going to test this branch on the courses we are having the issues with. Don't think it's ready for merge just yet without some real world testing. @nick-benoit14 says we are getting inconsistent results with this fix running on adhesion local.

lib/senkyoshi/canvas_course.rb
```diff
@@ -142,8 +142,17 @@ def upload_scorm_package(scorm_package, course_id, tmp_name)
...
-          result = JSON.parse(resp.body)["response"]
-          result["points_possible"] = scorm_package.points_possible
+          result = nil
+          begin
+            result = JSON.parse(resp.body)["response"]
+            raise Exceptions::BadResponse.new if result.nil?
+            result["points_possible"] = scorm_package.points_possible
+          rescue Exceptions::BadResponse, JSON::ParserError => e
+            STDERR.puts(
+              "Error: Invalid response from Scorm Manager: #{e.to_s[0..50]}",
```
bfcoder: is there an `e.messages` or `e.errors`?

lib/senkyoshi/canvas_course.rb
```diff
@@ -153,8 +162,8 @@ def upload_scorm_package(scorm_package, course_id, tmp_name)
     # Creates assignments from all previously uploaded scorm packages
     ##
     def create_scorm_assignments(scorm_packages, course_id, local)
-      scorm_packages.each do |pack|
-        create_scorm_assignment(pack, course_id, local)
+      scorm_packages.map do |pack|
```
bfcoder, line 165: consider writing specs that don't depend on the return array? something like `expect create_scorm_assignment to be called 5 times`?

nick-benoit14: I looked into it a bit this morning, and I can't quite figure out how to make it work in Minitest. Maybe you have more experience and a better idea of how it could work though. What I am running into is that I would have to mock entire CanvasCourse in order to have any expectation about how many times a particular method in the CanvasCourse was called.

jamis: As long as the method is explicit about its behavior, I don't think it's a problem for it to return the assignments it creates. Just make it clear (in a comment, for instance) that `#create_scorm_assignment` returns the assignment it creates, and `#create_scorm_assignments` returns an array of all created assignments.

The issue with the code as it is now (to me, anyway) is that the behavior feels too implicit, and so writing a test that depends on the return value of the method feels like you're depending on internal behavior.

srphillips:
I think this is going to be put off until we can test against SCORM Engine.  We should be getting a license Soon<sup>tm</sup>.

PR #73, opened 2017-02-16
Prevent `NoMethodError` when calling `xml_data.name`. [#139954609]

conradbeach:
We encountered a .dat file that had no XML and just said “This resource intentionally left blank”. Without the XML, Nokogiri can't parse anything out and `xml_data` gets set to `nil`. Senkyoshi throws an error on the next line when `.name` is called on `xml_data`.

This change fixes that bug by first checking that `xml_data` has data before getting the name from it. If `xml_data` is falsey, a default name of "empty" is used instead.


srphillips (approved):
Very simple fix with spec.  Looks good.

PR #74, opened 2017-02-16
Fix UTF-8, ASCII-8BIT Compatibility Error [#139953217]

conradbeach:
`Zip::Entry#name` returns an ASCII-8BIT encoded string. Instances of `SenkyoshiFile` had this encoding for the strings in `@xid` and `@path`. That had the potential to cause compatibility issues. This forces UTF-8 encoding on those instance variables.

It's worth noting that `Zip::Entry#name` is called elsewhere by Senkyoshi. It might be necessary to force UTF-8 encoding in other places as well. I chose just to fix the issue we were actually experiencing for now.

### DIFF
--- a/lib/senkyoshi/models/file.rb
+++ b/lib/senkyoshi/models/file.rb
@@ -10,10 +10,16 @@ module Senkyoshi
     ].freeze
 
     def initialize(zip_entry)
-      @path = strip_xid zip_entry.name.force_encoding("UTF-8")
+      begin
+        entry_name = zip_entry.name.encode("UTF-8")
+      rescue Encoding::UndefinedConversionError
+        entry_name = zip_entry.name.force_encoding("UTF-8")
+      end
+
+      @path = strip_xid entry_name
       @location = extract_file(zip_entry) # Location of file on local filesystem
 
-      base_name = File.basename(zip_entry.name.force_encoding("UTF-8"))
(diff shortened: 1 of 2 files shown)


conradbeach:
I looked into using `encode` instead of `force_encoding`, but it looks like `encode` is too gentle. On the example course, it errors out about a quarter of the time saying it can't perform the conversion. 

What I decided to do is try `encode` first and then use `force_encoding` after that if `encode` doesn't work. Thoughts?

srphillips:
Do we know which characters it's failing on specifically?

conradbeach:
`\xE2`

The full error message (for every error) is this:`<Encoding::UndefinedConversionError: "\xE2" from ASCII-8BIT to UTF-8>`

conradbeach:
We could specifically handle that character and covert it manually. `encode` gives you a way to do that. I think we'd still want to fall back on `force_encoding` though. This is just one course and there might be instances of other problem characters at some point.

I'm not sure trying to manually convert the character will be all that valuable though, `force_encoding` is already making a reasonable conversion.

srphillips:
Agreed that it might come up in other places, just curious about why it's still failing.  `force_encoding` is fine for now.

srphillips (approved):
Approved!

conradbeach:
Sounds good. Thanks.

PR #75, opened 2017-02-22
Remove unnecessary fileutils gem.

bfcoder:
This is already included with ruby and we don’t need to include it here.


blunckr:
sweet!

blunckr (approved):

PR #76, opened 2017-02-22
upgrade rake

blunckr:
canvas wants a new newer version of rake now


bfcoder (approved):
Looks good. This is going to make the rubygem gods angry, but that is ok with me. :]

blunckr:
I thought about doing `~> 12.0`, but that might not work with all our canvas installations.

PR #85, opened 2017-10-27
Update to allow customizing the oauth consumer key for scorm

bfcoder:
The problem right now with this gem, is the oauth consumer key has changed in Adhesion, so this is now broken. This allows you to set this value as you need it.


srphillips (approved):
Looks good. Thanks for making this configurable.  Go ahead and merge.

PR #87, opened 2017-10-30
Add support for authenticating with a JWT instead of a shared secret

bfcoder:


srphillips (approved):
This looks great, I approve!

PR #88, opened 2017-10-31
Fix SCORM manager integration

bfcoder:
Extract package_id from upload
Update scorm manager with data from lms after assignment is created
Bump version to 1.0.11

### DIFF
--- a/lib/senkyoshi/canvas_course.rb
+++ b/lib/senkyoshi/canvas_course.rb
@@ -135,10 +135,38 @@ module Senkyoshi
         assignment__points_possible__: scorm_package["points_possible"],
       }
 
-      CanvasCourse.client.create_assignment(
+      lms_assignment = CanvasCourse.client.create_assignment(
         course_id,
         scorm_package["title"],
         payload,
+      )
+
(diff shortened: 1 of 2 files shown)


lib/senkyoshi/canvas_course.rb
```diff
@@ -161,9 +189,11 @@ def upload_scorm_package(scorm_package, course_id, tmp_name)
           },
           Authorization: "Bearer #{AuthToken.issue_token}",
         ) do |resp|
-          result = JSON.parse(resp.body)["response"]
-          result["points_possible"] = scorm_package.points_possible
-          result
+          result = JSON.parse(resp.body)
+          response = result["response"]
```
srphillips, line 193: This is more descriptive of the process, but is there any other reason we are making `result` it's own variable if it's never used except to get the response out of it?

bfcoder: It is used on line 195.

srphillips: You are correct, my apologies!

srphillips (approved):
Looks good, thanks for fixing this @bfcoder

